PR #1, opened 2020-10-05
Update the OpenShift connector sidecar

ericwill:
Use odo 2.0.0 which is required for vscode-openshift-tooling 0.2.0

Part of CRW-1259

### DIFF
--- a/codeready-workspaces-plugin-openshift/Dockerfile
+++ b/codeready-workspaces-plugin-openshift/Dockerfile
@@ -15,7 +15,7 @@
 FROM ubi8-minimal:8.2-349
 
 ENV \
-    ODO_VERSION="v2.0.0" \
+    ODO_VERSION="v1.2.6" \
     KUBECTL_VERSION="v1.18.9" \
     HOME=/home/theia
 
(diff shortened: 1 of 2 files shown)


ericwill:
@nickboldt how do I get the image digest/tag to use in the updated `meta.yaml`? Or will it just be `registry.redhat.io/codeready-workspaces/plugin-openshift-rhel8:2.5`?

nickboldt (changes requested):
wrong file updated. this is generated from changes in get-sources-jenkins.sh script.

https://github.com/redhat-developer/codeready-workspaces-images/commit/ce5700bd23f5b26c3ec9840f445ff25e08da287e

then when the sync runs, you'll get Dockerfile and sources files updated:

https://github.com/redhat-developer/codeready-workspaces-images/commit/f229cc6c07d0ebc00c27f9929f5e7002419596a1

ericwill:
> should we update kubectl version too? What does odo require? Also do we need oc 4.5 ? we're on 4.4 right now.

Looks like we're good on that front -- upstream we're on kubectl 1.18.6, and oc 4.3.3 without any issue. odo is cool with 4.3.3 or greater

ericwill:
@nickboldt let's merge this one (if it's good to go) so I can test https://github.com/redhat-developer/codeready-workspaces/pull/352

PR #4, opened 2020-10-20
CRW-1298 - Upgrade default Kamel from 1.1.1 to 1.2.0

apupier:
it is the default used by VS Code Extension Tooling for Apache Camel K
0.0.18

nota: this is a blind commit. I haven't a local environment to test it.

is this file needs to be updated too?
How?
https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2.5-rhel-8/codeready-workspaces-plugin-kubernetes/sources

### DIFF
--- a/codeready-workspaces-plugin-kubernetes/Dockerfile
+++ b/codeready-workspaces-plugin-kubernetes/Dockerfile
@@ -16,7 +16,7 @@ FROM ubi8-minimal:8.2-349
 
 ENV \
     KUBECTL_VERSION="v1.18.9" \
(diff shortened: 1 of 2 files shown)


nickboldt:
Running the get-sources script will fetch new binaries, commit them to sources file, and update the Dockerfile at the same time.

nickboldt:
Building in https://codeready-workspaces-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CRW_CI/view/Pipelines_Next/job/crw-plugins-and-stacks_2.5/120/console

PR #7, opened 2020-10-30
exclude some files from the rsync (including...

nickboldt:
exclude some files from the rsync (including sources and .gitignore, as they're generated by rhpkg

Change-Id: I4e0cd05a5452abfcc45a9471fde1028a94239081
Signed-off-by: nickboldt <nboldt@redhat.com>


nickboldt:
Wrong solution - problem is that upstream builds were failing and we ended up with invalid tarballs.

PR #11, opened 2020-11-26
Move OpenShift connector sidecar back to using odo 1.2.6

ericwill:
Needed for redhat-developer/codeready-workspaces/pull/376

### DIFF
--- a/codeready-workspaces-plugin-openshift/Dockerfile
+++ b/codeready-workspaces-plugin-openshift/Dockerfile
@@ -15,7 +15,7 @@
 FROM ubi8-minimal:8.3-201
 
 ENV \
-    ODO_VERSION="v2.0.1" \
+    ODO_VERSION="v1.2.6" \
     KUBECTL_VERSION="v1.18.10" \
     HOME=/home/theia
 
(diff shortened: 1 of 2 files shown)


codeready-workspaces-plugin-openshift/Dockerfile
```diff
@@ -15,7 +15,7 @@
 FROM ubi8-minimal:8.3-201
 
 ENV \
-    ODO_VERSION="v2.0.1" \
+    ODO_VERSION="v1.2.6" \
```
nickboldt, line 18: need to set this in the get-sources script too, or else it could be overwritten. and also need to set it there so that sources file will be regenerated when built in brew.

nickboldt (changes requested):

nickboldt:
Fixed in https://github.com/redhat-developer/codeready-workspaces-images/commit/27621cdb12579bdff20ae516a30a9de2ebc65245 and https://github.com/redhat-developer/codeready-workspaces-images/commit/501ab729761893ed0222d0513af14cd4e7585e57

nickboldt:
Rebuilt from https://codeready-workspaces-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CRW_CI/view/Pipelines_Next/job/crw-plugins-and-stacks_2.5/152/

PR #12, opened 2020-12-04
Cleanup images with node

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

- Remove node from dotnet, php, cpp, golang images;
- Update node from 10 to 12 in kubernetes-plugin image

### DIFF
--- a/codeready-workspaces-stacks-golang/Dockerfile
+++ b/codeready-workspaces-stacks-golang/Dockerfile
@@ -49,15 +49,16 @@ COPY . /tmp/assets/
 # http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/layered/rhel8/s390x/rhocp/4.6/os/Packages/o/openshift-clients-4.6.0-202010081244.p0.git.3794.4743d24.el8.s390x.rpm
 # http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/layered/rhel8/ppc64le/rhocp/4.6/os/Packages/o/openshift-clients-4.6.0-202010081244.p0.git.3794.4743d24.el8.ppc64le.rpm
 RUN useradd -u 1000 -G wheel,root -d /home/jboss --shell /bin/bash -m jboss && \
+    dnf remove -y nodejs && \
     yum install -y \
     # next line are c/c++ compiler stuff which we apparently need for new golangci-lint
     make cmake gcc gcc-c++ glibc-devel zlib-devel libstdc++ libstdc++-devel \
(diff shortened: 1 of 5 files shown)


nickboldt:
Looks good! 

but why do we need node 12 in the k8s image at all?

ericwill:
> Looks good!
> 
> but why do we need node 12 in the k8s image at all?

I looked through the history for this image and couldn't find a reason why we needed it.

svor:
> > Looks good!
> > but why do we need node 12 in the k8s image at all?
> 
> I looked through the history for this image and couldn't find a reason why we needed it.

@nickboldt @ericwill I suspect node is needed for vscode-yaml https://github.com/svor/codeready-workspaces/blob/sv/testImages/dependencies/che-plugin-registry/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.2.0/meta.yaml#L20. 
@JPinkney can correct me if i'm wrong

JPinkney:
Yeah, it's needed for vscode-yaml

ericwill:
Of course it's needed for `vscode-yaml`, I even mentioned that during a standup this week :man_facepalming: sorry for the noise

codeready-workspaces-stacks-golang/Dockerfile
```diff
@@ -72,13 +73,11 @@ RUN useradd -u 1000 -G wheel,root -d /home/jboss --shell /bin/bash -m jboss && \
 ADD entrypoint.sh ${HOME}/entrypoint.sh
 RUN mkdir -p ${HOME}/go && \
     tar -xvf /tmp/assets/codeready-workspaces-stacks-language-servers-dependencies-golang-$(uname -m).tar.gz -C /tmp && \
-    cp -R /tmp/node_modules ${HOME}/che/ls-golang && \
```
nickboldt, line 75: So we don't need the node deps from codeready-workspaces-stacks-language-servers-dependencies-golang anymore? 

https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/crw-2.6-rhel-8/golang/build.sh#L41-L45

svor: at least I don't see any reason why we need that

nickboldt: Asked @kcrane about this one too and he can't recall any reason either. 

So we can probably merge this safely. :D

nickboldt (approved):

codeready-workspaces-stacks-golang/Dockerfile
nickboldt: It's strange... seems we copy node sources for the golang lang server into the side car but then don't actually compile it. Does the lang server get built on demand somewhere?

PR #21, opened 2021-03-08
CRW-1420 Providing lombok.jar for Java plugin to be activated from a devfile

sunix:
Providing lombok.jar to be activated from a devfile:

```yaml
  - id: redhat/java11/latest
    preferences:
      java.jdt.ls.vmargs: '-javaagent:/lombok.jar'
```

Reference issue: https://issues.redhat.com/browse/CRW-1420

Tested java11 plugin with this devfile in devsandbox (would only work in devsandbox as I have hardcoded the extension urls to devsandbox plugin registry):
```yaml
apiVersion: 1.0.0
metadata:
  generateName: java-lombok-
projects:
  - name: lombok-project-example
    source:
      location: 'https://github.com/chuucks/LOMBOK-PROJECT-EXAMPLE'
      type: git
      branch: master
components:
  - preferences:
      java.jdt.ls.vmargs: '-javaagent:/lombok.jar'
      java.server.launchMode: Standard
    type: chePlugin
    reference: 'https://gist.github.com/sunix/6a5b8da1f15390436b5faca775391278/raw/c04ca692d6d1dc2550c79e3483227bf0249d0d66/meta.yaml'
  - id: redhat/dependency-analytics/latest
    type: chePlugin
  - mountSources: true
    memoryLimit: 512Mi
    type: dockerimage
    volumes:
      - name: m2
        containerPath: /home/jboss/.m2
    alias: maven
    image: 'registry.redhat.io/codeready-workspaces/plugin-java11-rhel8@sha256:c13f0bb5f41dfe19ae90912392afc85aa8128cd3d4f7c01131cabe8d5ae3e87a'
    env:
      - value: '-XX:MaxRAMPercentage=50.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xms20m -Djava.security.egd=file:/dev/./urandom -Duser.home=/home/jboss'
        name: JAVA_OPTS
      - value: $(JAVA_OPTS)
        name: MAVEN_OPTS
commands:
  - name: maven build
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/lombok-project-example'
        type: exec
        command: mvn clean install
        component: maven

```

### DIFF
--- a/codeready-workspaces-plugin-java11-openj9/Dockerfile
+++ b/codeready-workspaces-plugin-java11-openj9/Dockerfile
@@ -17,6 +17,7 @@ USER root
 ENV \
     GRADLE_VERSION="6.1" \
     MAVEN_VERSION="3.6.3" \
+    LOMBOK_VERSION="1.18.18" \
     JAVA_HOME=/usr/lib/jvm/java-11-openj9 \
     PATH="/usr/lib/jvm/java-11-openj9:/opt/gradle/bin:/opt/apache-maven/bin:/usr/bin:${PATH:-/bin:/usr/bin}" \
     MANPATH="/usr/share/man:${MANPATH}" \
@@ -30,6 +31,7 @@ ENV \
 # run './get-sources-jenkins.sh --force-pull --nobuild' to generate this file, or 
 # use 'rhpkg sources' to fetch existing version from dist-git.
 COPY gradle-${GRADLE_VERSION}-bin.zip apache-maven-${MAVEN_VERSION}-bin.tar.gz /tmp/
+COPY lombok-${LOMBOK_VERSION}.jar /lombok.jar
 
 # NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io
 # COPY content_sets.repo /etc/yum.repos.d/
--- a/codeready-workspaces-plugin-java8-openj9/Dockerfile
+++ b/codeready-workspaces-plugin-java8-openj9/Dockerfile
@@ -17,6 +17,7 @@ USER root
 ENV HOME=/home/jboss \
     NODEJS_VERSION=12 \
     MAVEN_VERSION="3.6.3" \
+    LOMBOK_VERSION="1.18.18" \
     PYTHON_VERSION="3.8" \
     JAVA_HOME=/usr/lib/jvm/java-1.8.0-openj9 \
     PATH=$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/lib/jvm/java-1.8.0-openj9:/opt/apache-maven/bin:/usr/bin:$PATH \
@@ -28,6 +29,7 @@ ENV HOME=/home/jboss \
 
 # built in Brew, use get-sources-jenkins.sh to pull latest
 COPY . /tmp/assets/
+COPY lombok-${LOMBOK_VERSION}.jar /lombok.jar
 
 # NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io
 # COPY content_set*.repo /etc/yum.repos.d/
--- a/codeready-workspaces-plugin-java8/Dockerfile
+++ b/codeready-workspaces-plugin-java8/Dockerfile
@@ -17,6 +17,7 @@ USER root
 ENV HOME=/home/jboss \
     NODEJS_VERSION=12 \
     MAVEN_VERSION="3.6.3" \
+    LOMBOK_VERSION="1.18.18" \
     PYTHON_VERSION="3.8" \
     JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk \
     PATH=$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/lib/jvm/java-1.8.0-openjdk:/opt/apache-maven/bin:/usr/bin:$PATH \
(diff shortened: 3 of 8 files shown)


codeready-workspaces-plugin-java11-openj9/Dockerfile
```diff
@@ -81,6 +82,8 @@ RUN microdnf install -y bash tar gzip unzip which shadow-utils findutils wget cu
     gradle -v && \
     echo "========"
 
+RUN wget -O /lombok.jar https://projectlombok.org/downloads/lombok-${LOMBOK_VERSION}.jar
```
nickboldt, line 85: Can't fetch random stuff from the internet in a Brew build.

codeready-workspaces-plugin-java8-openj9/Dockerfile
```diff
@@ -130,6 +131,8 @@ RUN \
     pylint --version && \
     echo "========"
 
+RUN wget -O /lombok.jar https://projectlombok.org/downloads/lombok-${LOMBOK_VERSION}.jar
```
nickboldt, line 134: wget is not allowed in Brew builds. 

See this approach to unpacking a tarball that was previously built / fetched earlier, eg., for Maven 3.6.3

https://github.com/redhat-developer/codeready-workspaces-images/pull/21/files#diff-54b420cfaa5bf8472da078fefe1e4a579631fcaa91170574f78476d0cb87166fR84

If we don't trust the upstream provider (eg., bad license, no CVE process, see https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/crw-2-rhel-8/node10/build.sh or any of the other Dockerfile / build.sh scripts in that repo for creating new tarballs from upstream sources.

Then use get-sources script to fetch tarballs from Jenkins.

https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-plugin-java8/get-sources-jenkins.sh#L67-L68

nickboldt (changes requested):

codeready-workspaces-plugin-java8-openj9/Dockerfile
ericwill: > If we don't trust the upstream provider (eg., bad license, no CVE process, see https://github.com/redhat-developer/codeready-workspaces-deprecated/blob/crw-2-rhel-8/node10/build.sh or any of the other Dockerfile / build.sh scripts in that repo for creating new tarballs from upstream sources.
> 
> Then use get-sources script to fetch tarballs from Jenkins.
> 
> https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-plugin-java8/get-sources-jenkins.sh#L67-L68

[Lombok is on GitHub](https://github.com/rzwitserloot/lombok), and seem they have a fairly active community. They even have a [security policy](https://github.com/rzwitserloot/lombok/blob/master/SECURITY.md). I think we can trust it enough to not need to build from source -- just download the .jar and stick it into a tar archive, which can then be copied/unzipped as described above.

@nickboldt WDYT?

nickboldt: if we want to treat them like another lang server dependency like those in https://github.com/redhat-developer/codeready-workspaces-deprecated or a Maven- or Gradle-like mature project, I'm OK with that. I can add it to the ever-growing list of 'not from source' prodsec exceptions we maintain.

codeready-workspaces-plugin-java11-openj9/Dockerfile
nickboldt: please move this fetch into https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-plugin-java11/get-sources-jenkins.sh#L56-L57 and use curl

codeready-workspaces-plugin-java8-openj9/Dockerfile
nickboldt: apologies @sunix -- I steered you to the wrong place. Should have pointed you to https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-plugin-java11/get-sources-jenkins.sh#L56-L57

codeready-workspaces-plugin-java8/Dockerfile
```diff
@@ -24,10 +24,12 @@ ENV HOME=/home/jboss \
     JAVACONFDIRS="/etc/java${JAVACONFDIRS:+:}${JAVACONFDIRS:-}" \
     XDG_CONFIG_DIRS="/etc/xdg:${XDG_CONFIG_DIRS:-/etc/xdg}" \
     XDG_DATA_DIRS="/usr/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}" \
-    M2_HOME="/opt/apache-maven" 
+    M2_HOME="/opt/apache-maven" \
+    LOMBOK_VERSION="1.18.18"
```
nickboldt, line 28: could you include LOMBOK_VERSION in with the other _VERSION envs at the top? alphabetical order would be ideal :)

nickboldt (changes requested):
Please include the same fixes for the plugin-javaX-openj9 Dockerfiles and get-sources*.sh scripts.

nickboldt (approved):

PR #23, opened 2021-03-30
Fix codeready-workspaces-stacks-php local build

sunix:


nickboldt:
more complete fix generated by https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/product/getLatestRPM.sh

https://github.com/redhat-developer/codeready-workspaces-images/commit/a2822e8166cd94e6eb21256e7fff0e40546dd984

PR #32, opened 2021-05-12
Adapt to the new plugin registry

ericwill:
* Create and commit bootstrap.Dockerfile -- don't generate it with sed
* Inside bootstrap.Dockerfile we generate the plugin registry meta.yamls/artifacts without pushing an image
* Tar those up into a tar archive called resources.tgz
* Business as usual after that

Part of CRW-1648

### DIFF
--- a/codeready-workspaces-pluginregistry/bootstrap.Dockerfile
+++ b/codeready-workspaces-pluginregistry/bootstrap.Dockerfile
@@ -1,0 +1,37 @@
+#
+# Copyright (c) 2018-2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/python-38
+FROM ubi8-python-38:1-54.1618436884 as builder
+USER 0
+
+# Keep old BOOTSTRAP variable as the rhel.install script still checks them
+ARG BOOTSTRAP=true
+ENV BOOTSTRAP=${BOOTSTRAP}
+
+# Install deps
+COPY ./build/dockerfiles/content_sets_rhel8.repo /etc/yum.repos.d/
+COPY ./build/dockerfiles/rhel.install.sh /tmp
+RUN /tmp/rhel.install.sh && rm -f /tmp/rhel.install.sh
+
+# Copy files needed for the plugin registry build/artifact creation
+COPY ./build.sh ./*.yml ./*.yaml ./*.js ./*.json ./yarn.lock /build/
+COPY ./.yarn /build/.yarn/
+COPY ./tools/ /build/tools/
+COPY ./build /build/build/
+
+# Run plugin registry build to generate artifacts
+WORKDIR /build/
+RUN SKIP_FORMAT=true SKIP_LINT=true SKIP_TEST=true ./build.sh --offline --skip-oci-image --skip-digest-generation
+
+# Archive artifacts to be copied out by Jenkins
+RUN tar -czvf resources.tgz ./output/v3/ && mkdir /tmp/resources/ && cp ./resources.tgz /tmp/resources/
--- a/codeready-workspaces-pluginregistry/build/scripts/sync.sh
+++ b/codeready-workspaces-pluginregistry/build/scripts/sync.sh
@@ -93,18 +93,7 @@ sed "${TARGETDIR}/build/dockerfiles/Dockerfile" --regexp-extended \
     -e 's|# (COPY root-local.tgz)|\1|' \
     `# only enable rhel8 here -- don't want centos or epel ` \
     -e 's|^ *(COPY .*)/content_set.*repo (.+)|\1/content_sets_rhel8.repo \2|' \
-    `# Comment out PATCHED_* args from build and disable update_devfile_patched_image_tags.sh` \
-    -e 's|^ *ARG PATCHED.*|# &|' \
-    -e '/^ *RUN TAG/,+3 s|.*|# &| ' \
-    `# Disable intermediate build targets` \
-    -e 's|^ *FROM registry AS offline-registry|# &|' \
-    -e '/^ *FROM builder AS offline-builder/,+3 s|.*|# &|' \
-    -e 's|^[^#]*--from=offline-builder.*|# &|' \
-    `# Enable cache_artifacts.sh` \
-    -e '\|swap_images.sh|i # Cache projects in CRW \
-COPY ./build/dockerfiles/rhel.cache_artifacts.sh resources.tgz /tmp/ \
-RUN /tmp/rhel.cache_artifacts.sh /build/v3/ && rm -rf /tmp/rhel.cache_artifacts.sh /tmp/resources.tgz \
-' > "${TARGETDIR}/Dockerfile"
+  > "${TARGETDIR}/Dockerfile"
 
 cat << EOT >> "${TARGETDIR}/Dockerfile"
 ENV SUMMARY="Red Hat CodeReady Workspaces pluginregistry container" \\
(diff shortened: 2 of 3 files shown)


codeready-workspaces-pluginregistry/get-sources-jenkins.sh
```diff
@@ -32,17 +31,11 @@ function log()
...
-  -e 's|COPY (.*) resources.tgz (.*)|COPY \1 \2|' \
-  -e 's|ARG BOOTSTRAP=.*|ARG BOOTSTRAP=true|' \
-  -e 's|ARG USE_DIGESTS=.*|ARG USE_DIGESTS=false|' \
-  -e 's|^ *COPY root-local.tgz|# &|' \
+sed bootstrap.Dockerfile -i --regexp-extended \
   `# replace org/container:tag with reg-proxy/rh-osbs/org-container:tag` \
   -e "s#^FROM ([^/:]+)/([^/:]+):([^/:]+)#FROM registry-proxy.engineering.redhat.com/rh-osbs/\1-\2:\3#" \
   `# replace ubi8-minimal:tag with reg-proxy/rh-osbs/ubi-minimal:tag` \
-  -e "s#^FROM ([^/:]+):([^/:]+)#FROM registry-proxy.engineering.redhat.com/rh-osbs/\1:\2#" \
-  -e 's|# (COPY .*content_sets.*)|\1|' \
```
nickboldt, line 44: you might want this one back... we want content_sets*.repo enabled for bootstrap build but NOT for the brew build.

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
```diff
@@ -77,8 +70,7 @@ sudo rm -fr ${tmpDir} ${BEFORE_DIR}
 
 # resources.tgz
 tmpDir=$(mktemp -d)
-docker run --rm -v ${tmpDir}/:/tmp/resources/ --entrypoint /bin/bash ${tmpContainer} -c \
-  "cd /build && cp -r ${filesToInclude} /tmp/resources/"
+docker run --rm -v ${tmpDir}/:/tmp/resources/
```
nickboldt, line 73: if you just run the container, will it automatically copy everything from /build into /tmp/resources? Or do you need to manually copy stuff out of /build ?

ericwill: I'm checking that now, haven't tried it yet

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
ericwill: noted

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
nickboldt: ah, I see `cp ./resources.tgz /tmp/resources/` which ill put resources.tgz into ${tmpDir}/ so we can commit it w/ rhpkg

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
ericwill: Since bootstrap.Dockerfile will never be run in Brew I don't think this is an issue, actually

codeready-workspaces-pluginregistry/bootstrap.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/python-38
+FROM registry-proxy.engineering.redhat.com/rh-osbs/ubi8-python-38:1-54.1618436884 as builder
```
nickboldt, line 14: since this is running in Jenkins, it could just be the latest public image from registry.redhat.io or registry.access.redhat.com

codeready-workspaces-pluginregistry/build/scripts/sync.sh
```diff
@@ -93,18 +93,7 @@ sed "${TARGETDIR}/build/dockerfiles/Dockerfile" --regexp-extended \
     -e 's|# (COPY root-local.tgz)|\1|' \
     `# only enable rhel8 here -- don't want centos or epel ` \
     -e 's|^ *(COPY .*)/content_set.*repo (.+)|\1/content_sets_rhel8.repo \2|' \
```
nickboldt, line 95: we don't need/want .repo files enabled in Brew.

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
```diff
@@ -31,32 +29,27 @@ function log()
...
+  -e "s#^FROM ([^/:]+):([^/:]+)#FROM registry-proxy.engineering.redhat.com/rh-osbs/\1:\2#"
 echo "======= BOOTSTRAP DOCKERFILE =======>"
 cat bootstrap.Dockerfile
 echo "<======= BOOTSTRAP DOCKERFILE ======="
 echo "======= START BOOTSTRAP BUILD =======>"
-# do not want digests in the BOOTSTRAP build so override default with false
+
 docker build -t ${tmpContainer} . --no-cache -f bootstrap.Dockerfile \
-  --target builder --build-arg BOOTSTRAP=true --build-arg USE_DIGESTS=false
+  --target builder --build-arg BOOTSTRAP=true
```
nickboldt, line 44: is USE_DIGESTS=false by default?

codeready-workspaces-pluginregistry/bootstrap.Dockerfile
ericwill: Sorry I committed the sed'd changes, it should just be `ubi8-python-38:1-54.1618436884`

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
ericwill: the bootstrap.Dockerfile doesn't care about digests anymore, that's handled in the plugin registry Dockerfile now (along with the memory limit swapping and such)

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
```diff
@@ -75,30 +68,14 @@ if [[ ${TAR_DIFF} ]]; then
...
-  echo "${TAR_DIFF2}"
-  echo "***** END DIFF"
-  pushd ${tmpDir} >/dev/null && tar czf resources.tgz ./* && popd >/dev/null
-  mv -f ${tmpDir}/resources.tgz .
-fi
-sudo rm -fr ${tmpDir}
-rm bootstrap.Dockerfile
+# we always need a fresh resources.tgz to guarantee proper timestamps and latest vsix files
+rm -f ./resources.tgz
+docker create --name pluginregistryBuilder ${tmpContainer}
```
nickboldt, line 73: thee other approach is to run the container against a mounted volume and copy out the file / create a tarball on the fly... but I like this appoach as long as it works for both docker & podman. 

(ideally this script would use podman as preference and fall back to docker when not available)

codeready-workspaces-pluginregistry/bootstrap.Dockerfile
nickboldt: plsfix kthxbai.

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
nickboldt: ok, that works.

codeready-workspaces-pluginregistry/build/scripts/sync.sh
ericwill: are you sure? I didn't introduce this change it was there before

codeready-workspaces-pluginregistry/get-sources-jenkins.sh
ericwill: Everything here will work with podman (I tested using both)

codeready-workspaces-pluginregistry/build/scripts/sync.sh
nickboldt: might be that it was in the file but `#`'d out... and then transformed into the bootstrap flavour by uncommenting it.

nickboldt: see https://github.com/redhat-developer/codeready-workspaces-images/pull/32#discussion_r631474254 above where it's uncommented.

PR #53, opened 2021-07-09
Adapt CRW dashboard to a newer product.json.template & update request feature link

sleshchenko:
Adapt CRW dashboard to a newer product.json.template from https://github.com/eclipse-che/che-dashboard/pull/294
+ update request feature link

It should fix https://issues.redhat.com/browse/CRW-1879


nickboldt (approved):
seems reasonable; assuming this is only for 2.x / 2.11 (not to be backported to 2.10)

PR #55, opened 2021-07-12
Fix typo in the dashboard synchronization script

akurinnoy:
Fixed typo `--ignore-engine` => `--ignore-engines`

### DIFF
--- a/codeready-workspaces-dashboard/build/scripts/sync.sh
+++ b/codeready-workspaces-dashboard/build/scripts/sync.sh
@@ -126,8 +126,8 @@ LABEL summary="\$SUMMARY" \\
 EOT
 
 # Patch rhel.Dockerfile to switch to yarn1
-sed -r -i \
-  -e 's|(RUN /dashboard/.yarn/releases/yarn-\*\.cjs install)|\1 --ignore-engine|' \
+sed -r -i '' \
+  -e 's|(RUN /dashboard/.yarn/releases/yarn-\*\.cjs install)|\1 --ignore-engines|' \
   ${TARGETDIR}/build/dockerfiles/rhel.Dockerfile
 
 echo "Converted Dockerfile"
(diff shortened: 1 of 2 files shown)


sleshchenko (approved):

codeready-workspaces-dashboard/build/scripts/sync.sh
```diff
@@ -126,8 +126,8 @@ LABEL summary="\$SUMMARY" \\
 EOT
 
 # Patch rhel.Dockerfile to switch to yarn1
-sed -r -i \
-  -e 's|(RUN /dashboard/.yarn/releases/yarn-\*\.cjs install)|\1 --ignore-engine|' \
+sed -r -i '' \
```
sleshchenko, line 129: what is '' ?

akurinnoy: -i means --in-place, and it expects a backup file extension. Empty string makes it not creating any backup files.

nickboldt: The `-i ''` only works on MacOS. On linux this will fail; you want simply `-i` on linux. 

```
$➔ echo "foo" > bar; cat bar; sed -i '' -r -e "s/o/a/g" bar; echo $?; cat bar
foo
sed: can't read : No such file or directory
2 << error code greater than 0 = bash failure if this is run in a process with `-e`
faa
```
vs.

```
$➔ echo "foo" > bar; cat bar; sed -i -r -e "s/o/a/g" bar; echo $?; cat bar
foo
0
faa
```

nickboldt:
and voila,
```
09:30:53 sed: can't read : No such file or directory
```

-- https://main-jenkins-csb-crwqe.apps.ocp4.prod.psi.redhat.com/job/CRW_CI/job/crw-dashboard_2.x/62/console

PR #56, opened 2021-07-13
Ignore engines while building the dashboard

akurinnoy:


sleshchenko (approved):
You can't just simply do CRW container build from the first try
image

One more try

PR #58, opened 2021-07-13
Fix dashboard build

akurinnoy:
Apart from using `lerna` to avoid incompatible engine issues, we also need to cache `node_modules` from each package.


sleshchenko (approved):

sleshchenko:
is bulding in https://main-jenkins-csb-crwqe.apps.ocp4.prod.psi.redhat.com/job/CRW_CI/job/crw-dashboard_2.x/68/

PR #64, opened 2021-07-23
chore(build): Add scripts to build images using github action

benoitf:
### What does this PR do?

Add a script to build either all images, a given set of images or only the images changed in the last commit

```
./build.sh --help
./build.sh : will build only modified folders since the last commit
./build.sh --all-images: rebuild all images : will build only modified folders since the last commit
./build.sh --folders=path1,path2: rebuild only those paths
```

### What issues does this PR fix or reference?
https://issues.redhat.com/browse/CRW-2081

### How to test this PR?
Log in to registry.redhat.io and then:

```
./.github/build.sh --folders=codeready-workspaces-traefik
```

### DIFF
--- a/.github/build.sh
+++ b/.github/build.sh
@@ -1,0 +1,226 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+set -e
+set -o pipefail
+
+help() {
+    echo "Usage: build.sh [options]"   
+    echo "--help: display this help"
+    echo "empty options: build only images with folders being changed by last commit. Can be overrided by:"
+    echo "--all-images: rebuild all images"
+    echo "--folders=path1,path2: rebuild only those paths"
+    echo "--arch=archName: to specify the architecture"
+}
+
+# initialize global variables
+init() {
+  BUILD_ALL_IMAGES=false
+  SPECIFIC_FOLDERS=""
+  ARCH="x86_64"
+}
+
+# parse arguments
+# FILTERED_ARGS are the arguments that are passed to the script minus the one being handled
+parse() {
+  while [ $# -gt 0 ]; do
+    case $1 in
+      --all-images)
+        BUILD_ALL_IMAGES=true
+        shift;;
+      --folders=*)
+        SPECIFIC_FOLDERS="${1#*=}"
+        shift ;;
+      --arch=*)
+        ARCH="${1#*=}"
+        shift ;;
+      --help)
+        help
+        exit 0;;
+      *)
+      shift;;
+    esac
+  done
+}
+
+# grab current version (short)
+# 2.14
+get_current_version() {
+  jq -r '.CRW_VERSION' VERSION.json
+} 
+
+# 2.14.0
+get_current_version_metdata() {
+  jq -r '.CSV_VERSION_METADATA' VERSION.json
+} 
+
+
+# get github release for the current version
+# param: $1 - artifact name (like 2.14.0-configbump-assets, etc)
+get_github_release() {
+    local -r assetReleaseName=$1
+    # need to call github API
+    # https://docs.github.com/en/rest/reference/repos#get-a-release-by-tag-name
+    curl -s -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/redhat-developer/codeready-workspaces-images/releases/tags/${assetReleaseName}"
+}
+
+# extract assets from a github release object
+# param: content of github release JSON
+extract_github_release_assets() {
+    jq -n "${1}" | jq -cr '.assets[] | {name:.name, url:.browser_download_url}'
+}
+
+# Get folder names of the images to build from the current git branch
+# no param
+get_folder_images_to_build() {
+    
+    # if variables BUILD_ALL_IMAGES is true then we build all folders
+    if [ "${BUILD_ALL_IMAGES}" = true ]; then
+        # get all folders
+        find . -type d -maxdepth 1 | grep codeready- | sed 's/^\.\///g' | sort -u
+    elif [ -n "${SPECIFIC_FOLDERS}" ]; then
+        # use folders that are specified
+        IFS=',' read -r -a tmpFoldersArray <<< "${SPECIFIC_FOLDERS}"
+        # wrap to a multi-line string
+        IFS=$'\n'; echo "${tmpFoldersArray[*]}"
+    else
+      # need to check which folders have been updated by this commit
+      # grab all modified files then, get the folder and then only report unique folders
+      git diff --name-only --diff-filter=d -r HEAD HEAD^1 |  cut -d "/" -f1 | uniq
+    fi
+}
+
+# Download assets for a given folder
+# param: $1 is the path to the folder to build
+# param: $2 is the name of the image
+# param: $3 are the assets
+download_build_assets() {
+    local -r folder=${1}
+    local -r imageName=${2}
+    local -r assets=${3}
+    
+    echo "Downloading assets for image ${imageName}..."
+
+    # iterate on each assets
+    local filename;
+    local downloadURL;
+    for asset in $assets; do
+
+        # name
+        filename=$(jq -n "${asset}" | jq -r '.name')
+
+        # file is not for this arch, skip it
+        if [[ "$filename" != *"${ARCH}.tar.gz"* ]]; then
+          echo "skipping ${filename} as not matching ${ARCH} arch"
+          continue;
+        fi
+
+        downloadURL=$(jq -n "${asset}" | jq -r '.url')
+
+        # check download URL is there else throw an error
+        if [ -z "${downloadURL}" ]; then
+            echo "Unable to find download URL for ${filename} required for image ${imageName} in folder ${folder}"    
+            exit 1
+        fi
+
+        # Download the file with curl
+        echo " - ${filename}..."
+        curl -s -L -o "${folder}/${filename}" "${downloadURL}"
+    done <<< "$(cat ./"${folder}"/sources)"
+}
+
+# Patch dockerfile content by adding registry prefix
+# param: $1 is the path of the dockerfile to patch
+patch_dockerfile_content() {
+    local -r dockerfilePath=${1}
+
+    # Patch Dockerfile to add redhat registry
+    sed "s/FROM rhel8/FROM registry.redhat.io\/rhel8/" "${dockerfilePath}" | \
+    sed "s/FROM ubi8/FROM registry.redhat.io\/ubi8/"
+}
+
+# Build the image for the given folder
+# param: $1 is the path to the folder to build
+# param: $2 is the name of the image
+build_image() {
+    local -r folder=${1}
+    local -r imageName=${2}
+
+    # the asset tag is composed of version and component name
+    local -r componentName="${folder//codeready-workspaces-/}"
+
+    local -r assetReleaseName="${currentVersion}-${componentName}-assets"
+
+    # get content of the github release for the given asset
+    echo "Fetching github metadata for ${componentName} with release ${assetReleaseName}..."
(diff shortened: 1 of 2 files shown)


.ci/build.sh
```diff
@@ -0,0 +1,230 @@
...
+    done <<< "$(cat ./"${folder}"/sources)"
+}
+
+# Patch dockerfile content by adding registry prefix
+# param: $1 is the path of the dockerfile to patch
+patch_dockerfile_content() {
+    local -r dockerfilePath=${1}
+
+    # Patch Dockerfile to add redhat registry
+    sed "s/FROM rhel8/FROM registry.redhat.io\/rhel8/" "${dockerfilePath}" | \
```
nickboldt, line 151: this might not work -- some images use public ubi8 org, but internally use rhel8 org for container references. 

So... might need more complicated rules here for the go-toolset base images.

benoitf: could you be more specific ?

for example machine-exec is using `FROM rhel8/go-toolset:1.15.13-4 as builder` and it's replaced by `FROM registry.redhat.io/rhel8/go-toolset:1.15.13-4 as builder` and then it works

nickboldt:
I see two "fixme" items in there -- can you resolve those before merging?

Also, I see we're using .github for the action and .ci for the script -- would it make sense to have everything in the .github folder, since the .ci folder convention seems to be going away ... we don't use it in Che any more for centos CI stuff, and we've renamed builds from ":ci" to ":next" ...

benoitf:
> Also, I see we're using .github for the action and .ci for the script -- would it make sense to have everything in the .github folder, since the .ci folder convention seems to be going away ... we don't use it in Che any more for centos CI stuff, and we've renamed builds from ":ci" to ":next" ...

I was wondering if we could have the script at the root folder.
As there is nothing related to github action in this file. but let's keep it for now 'internally in github folder" 👍

benoitf:
> I see two "fixme" items in there -- can you resolve those before merging?
One is an enhancement, I've renamed fixme

the other fixme is related to  https://issues.redhat.com/browse/CRW-2077 so I added a link to it

benoitf:
it should be good for step 1 now :-)

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+        BUILD_ALL_IMAGES=true
+        shift;;
+      --folders=*)
+        SPECIFIC_FOLDERS="${1#*=}"
+        shift ;;
+      --help)
+        help
+        exit 0;;
+      *)
+        FILTERED_ARGS="${FILTERED_ARGS} $1"
```
nickboldt, line 42: why are you collecting unused args? doesn't seem like you're doing anything with these. I'm sure the shellcheck gods would be angry about this. :)

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+      shift;;
+    esac
+  done
+}
+
+# grab current version
+get_current_version() {
+  # get the current version from the VERSION file of codeready-workspaces
+  # could look at a file inside the repository once https://issues.redhat.com/browse/CRW-2077 is done
+  curl -s https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/HEAD/dependencies/VERSION
```
nickboldt, line 52: I generally curl with -sSL in case of errors and links/redirections... also does HEAD always map to the default branch's latest commit? If so, this is a problem since I imagine at some point you'll want to run this script from the crw-2.y-rhel-8 stable branch too, not just the nightly/CI/next branch crw-2-rhel-8. 

Would it be better to compute the correct VERSION file from the current branch, eg., with `git rev-parse --abbrev-ref HEAD` ?

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+    # filename@SHa512.extension so we need to rebuild this
+    local -r regexpValue=$(echo "${expectedFilename}" | sed -rn 's/([^\.]*)([^.]*)/\1|\2/p')
+
+    IFS="|" read -r filenameWithoutExtension fileExtension  <<< "${regexpValue}"
+
+    # compute the link
+    search_filename=${filenameWithoutExtension}@${expectedSha512}${fileExtension}
+
+    # find the matching asset
+    jq --arg SEARCH_FILENAME "${search_filename}" -r '. | select(.name==$SEARCH_FILENAME) | .browser_download_url' <<< "${assetJsonContent}"
```
nickboldt, line 88: this likely needs a shellcheck disable on it, so it doesn't complain about `$SEARCH_FILENAME` not being in ""

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+
+# Patch dockerfile content by adding registry prefix
+# param: $1 is the path of the dockerfile to patch
+patch_dockerfile_content() {
+    local -r dockerfilePath=${1}
+
+    # Patch Dockerfile to add redhat registry
+    sed "s/FROM rhel8/FROM registry.redhat.io\/rhel8/" "${dockerfilePath}" | \
+    sed "s/FROM ubi8/FROM registry.redhat.io\/ubi8/"
+
```
nickboldt, line 154: empty line

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+}
+
+# initialize
+init
+
+# analyze args
+parse "$@"
+
+# build images, with either:
```
nickboldt, line 227: to avoid writing the same instructions/usage in 2 places, why not just mention the `parse()` and `help()` methods above?

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,50 @@
...
+name: Image Build
+on:
+  push:
+    branches:
+      - crw-2-rhel-8
+
+  workflow_dispatch:
+    inputs:
+      arguments:
+        description: 'CLI Argument'
```
nickboldt, line 19: A better description might be "parameters for the build" with examples like 
```
--all-images or --folders=codeready-workspaces-machineexec,codeready-workspaces-imagepuller,...
```

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,50 @@
...
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Clone source code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+#    - name: Login to quay.io
```
nickboldt, line 33: Does this need a `TODO: enable this when ready to push images to quay` ?

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+  local -r numberOfImages=$(echo "${foldersToBuild}" | wc -l | bc)
+  echo " => Found ${numberOfImages} images to build "
+
+  echo ""
+  local imageBuildCount=1
+  while IFS= read -r folderToBuild ; do
+    local imageName="quay.io/crw/${folderToBuild}:gh-${currentVersion}"
+    echo "${imageBuildCount} - Building image ${imageName} in the folder ${folderToBuild}..."
+    build_image "${folderToBuild}" "${imageName}" "${availableAssets}"
+    echo "Image ${imageName} successfully build."
```
nickboldt, line 212: s/build/built/

Also, is this where a `TODO: push multiarch images to quay.io/crw/...` should be?

.github/build.sh
```diff
@@ -0,0 +1,231 @@
...
+  echo "Extracting available assets..."
+  local -r availableAssets=$(extract_github_release_assets "${content}")
+
+  # Get list of folders where we need to run the build
+  echo "Computing list of folders to build..."
+  local -r foldersToBuild=$(get_folder_images_to_build)
+  
+  # if foldersToBuild is empty
+  if [ -z "${foldersToBuild}" ]; then
+    echo " => No image to build, skipping"
```
nickboldt, line 199: error message should clarify that none of the 3 conditions have been met:
* all images flag
* list of specific images via flag
* local/recent commits

.github/build.sh
nickboldt: also you don't mention that we can run this with no params and it should look for recent changes to build.

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,50 @@
...
+#      with:
+#        registry: quay.io
+#        username: ${{ secrets.QUAY_USERNAME }}
+#        password: ${{ secrets.QUAY_PASSWORD }}
+
+    - name: Login to registry.redhat.io
+      uses: docker/login-action@v1
+      with:
+        registry: registry.redhat.io
+        username: ${{ secrets.REGISTRY_REDHAT_IO_USERNAME }}
```
nickboldt, line 44: dumb question... did you already upload the crw-bot secret so this will work?

.github/build.sh
benoitf: shellcheck is passing. You can verify on shellcheck.net

.github/build.sh
benoitf: the would it better is the fix of https://issues.redhat.com/browse/CRW-2077 (the VERSION file in this repository)
I would not investigate more there.

And as it's only applied on default branch, HEAD is working.

.github/build.sh
benoitf: I don't have any shellcheck reports so what shellcheck are you using ?

benoitf: $SEARCH_FILENAME is a jq variable and we properly give the assignement in `--arg SEARCH_FILENAME "${search_filename}"`

and all jq parameter is wrapped into simple quotes to not interpret any variable so is it you that is suspicious but not shellcheck ?

.github/build.sh
benoitf: well for now there is no push so we don't need TODO.
it requires all assets to be there before doing the next step.

.github/build.sh
benoitf: > also you don't mention that we can run this with no params and it should look for recent changes to build.

it's written there: only images that are modified

.github/workflows/build-image.yaml
benoitf: yes and it works (I tested in my fork as well)

.github/build.sh
benoitf: (in any case I've removed the variable but it was not complaining)

.github/build.sh
benoitf: deleted comments and added a see usage()

.github/build.sh
benoitf: added a 'mode' echo before

nickboldt:
TODO before merging:
* generate VERSION file in images repo
* publish via automated process all the assets currently in pkgs.devel to GH, so that this process can be enabled

benoitf:
Updated the script as sha512 are not used anymore

it's still missing the CRW version in this repository / https://issues.redhat.com/browse/CRW-2077

Tested on traefik folder/x64

nickboldt:
Blocked by:
* GH release asset IDs should match their container names - https://issues.redhat.com/browse/CRW-2436
* CSV versions in job-config.json are invalid for newly added versions (2.14) and for 2.x - https://issues.redhat.com/browse/CRW-2435
* Add version of CodeReady workspaces in codeready-workspaces-images repository - https://issues.redhat.com/browse/CRW-2077

benoitf:
https://issues.redhat.com/browse/CRW-2436 is not a blocker (I have filtered out `stacks` and `plugin`) so AFAIK it could be merged

.github/build.sh
```diff
@@ -0,0 +1,226 @@
...
+    done <<< "$(cat ./"${folder}"/sources)"
+}
+
+# Patch dockerfile content by adding registry prefix
+# param: $1 is the path of the dockerfile to patch
+patch_dockerfile_content() {
+    local -r dockerfilePath=${1}
+
+    # Patch Dockerfile to add redhat registry
+    sed "s/FROM rhel8/FROM registry.redhat.io\/rhel8/" "${dockerfilePath}" | \
```
nickboldt, line 144: this could be done as a single sed and actually save the change to the file...
```
sed -r -i "${dockerfilePath}" -e  "s#FROM (rhel8|ubi8)#FROM registry.redhat.io/(\1)#g"
```

benoitf: yes but -i is not macos friendly

nickboldt: that's why you wrote sed_in_place for theia, right? :)

PR #65, opened 2021-07-26
Projector-based docker image for CRW

vzhukovs:
Introduce Projector-based docker image for CRW.

Current changes proposal provides synchronisation and shell scripts for building docker image for IntelliJ IDEA Community Edition 2020.3.3.


codeready-workspaces-idea/build/scripts/sync.sh
```diff
@@ -0,0 +1,133 @@
...
+make-release.sh
+" > /tmp/rsync-excludes
+echo "Rsync ${SOURCEDIR} to ${TARGETDIR}"
+rsync -azrlt --checksum --exclude-from /tmp/rsync-excludes --delete "${SOURCEDIR}"/ "${TARGETDIR}"/
+rm -f /tmp/rsync-excludes
+
+# ensure shell scripts are executable
+find "${TARGETDIR}"/ -name "*.sh" -exec chmod +x {} \;
+
+apply_sed() {
```
nickboldt, line 74: please use sed_in_place function from https://github.com/redhat-developer/codeready-workspaces-theia/blob/crw-2-rhel-8/build.sh#L199-L206 so we're consistent in our MacOS hackarounds. :D

codeready-workspaces-idea/build/scripts/sync.sh
```diff
@@ -0,0 +1,133 @@
...
+  SHORT_UNAME=$(uname -s)
+  if [ "$(uname)" == "Darwin" ]; then
+    sed -i '' "$1" "$2"
+  elif [ "${SHORT_UNAME:0:5}" == "Linux" ]; then
+    sed -i "$1" "$2"
+  fi
+}
+
+# Replace ubi8 with rhel8 version
+apply_sed "s/FROM registry.redhat.io\//FROM /" "${TARGETDIR}"/Dockerfile
```
nickboldt, line 84: using sed_in_place you can chain all these changes together in a single command, rather than 6:

https://github.com/redhat-developer/codeready-workspaces-theia/blob/crw-2-rhel-8/build.sh#L470-L481
https://github.com/redhat-developer/codeready-workspaces-theia/blob/crw-2-rhel-8/build.sh#L199-L206

codeready-workspaces-idea/get-sources.sh
```diff
@@ -0,0 +1,98 @@
...
+  if [[ ! -f "build/projector-server-assembly" ]]; then
+    log "[ERROR] 'build/projector-server-assembly' not found, so nothing to build."
+    exit 1;
+  else
+    mv build/projector-server-assembly projector-server-assembly
+  fi
+
+  rm -rf build
+
+  outputFiles="ide-packaging projector-server-assembly static"
```
nickboldt, line 57: where does static folder come from?

codeready-workspaces-idea/build/scripts/sync.sh
nickboldt: fixed in https://github.com/redhat-developer/codeready-workspaces-images/pull/66

codeready-workspaces-idea/build/scripts/sync.sh
nickboldt: fixed in https://github.com/redhat-developer/codeready-workspaces-images/pull/66

codeready-workspaces-idea/get-sources.sh
vzhukovs: Yep, I've created PR in upstream: https://github.com/che-incubator/jetbrains-editor-images/pull/43
Which replaces `ADD` instructions with `COPY` for less manipulations with Dockerfile in `get-sources.sh` in midstream. Also archived `static/` folder and provide it to Docker image as gzip archive.

>Are any of the generated files in ide-packaging or projector-server-assembly dependent on rpms/go/node code that's arch-specific?

They are doesn't use go. So there's no problem with it. Projector Client builds in `get-sources.sh` and it uses `kotlin-react-*` libraries. Need to check how the build will be performed in different archs. IntelliJ IDEA itself requires: `libsecret`, AFAIK, there are some problems on ppc64 with this package. Also need to check. Also IDEA itself uses: `libXext libXrender libXtst libXi libX11-xcb mesa-libgbm libdrm freetype` for own needs. Haven't check them on different archs yet.

nickboldt: libsecret is different on different arches. See https://github.com/redhat-developer/codeready-workspaces-theia/commit/474bbce5d1ab700b38e034da989d8a0e7570eae2 for how to handle installation on s390x and ppc64le for building bootstrap builds in Jenkins.

In OSBS/Brew, we only need this, as libsecret-devel is available from internal registries:

https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2.10-rhel-8/codeready-workspaces-theia-dev/Dockerfile#L22

PR #71, opened 2021-07-29
Install JDK as requirement for building Projector sources

vzhukovs:
Projector sources build using Gradle Wrapper by calling:
```
./gradlew :projector-server:distZip
```
under the hood of `./projector.sh` shell script. As a requirement, JDK 11 should be installed and `JAVA_HOME` env variable should be configured.


nickboldt:
related to https://issues.redhat.com/browse/CRW-1587

PR #78, opened 2021-08-05
chore: don't modify pvcStrategy

tolusha:

The default pvcStrategy is common in upstream.
It should be the same for CRW.


nickboldt:
added second commit to update copy of sync script in other folder; cherrypicked to 2.11 branch

PR #96, opened 2021-09-08
CRW-2196

SDawley:
Adding logic to check if the previous CSV version in job-config exists and using the version from the latest release if it isn't found.

### DIFF
--- a/codeready-workspaces-operator-metadata/build/scripts/sync.sh
+++ b/codeready-workspaces-operator-metadata/build/scripts/sync.sh
@@ -21,7 +21,7 @@ CRW_VERSION=${CSV_VERSION%.*} # tag 2.y
 UPSTM_NAME="operator"
 MIDSTM_NAME="operator-metadata"
 
-CSV_VERSION_PREV=""
+CSV_VERSION_PREV="2.10.1"
 
 usage () {
     echo "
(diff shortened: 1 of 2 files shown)


nickboldt (approved):
+1, tested locally and did everything I thought it would :) 

```
$➔ ./build/scripts/sync.sh -v 2.12.0 -s ~/0/op -t `pwd` 
FATA[0001] Error parsing image name "docker://registry.redhat.io/codeready-workspaces/crw-2-rhel8-operator-metadata:2.11.0": Error reading manifest 2.11.0 in registry.redhat.io/codeready-workspaces/crw-2-rhel8-operator-metadata: error parsing HTTP 404 response body: invalid character 'N' looking for beginning of value: "Not found\n" 
[INFO] Using local registry.redhat.io/codeready-workspaces/crw-2-rhel8-operator-metadata:latest (6e8e808ae060)...
[INFO] Extract from container () ...
[INFO] Container registry.redhat.io/codeready-workspaces/crw-2-rhel8-operator-metadata:latest (6e8e808ae060) unpacked to /tmp/registry.redhat.io-codeready-workspaces-crw-2-rhel8-operator-metadata-latest-1631279730
[INFO] For 2.12 / crw-2-rhel-8:
[INFO]   CSV_VERSION_PREV   = 2.10.1
```

Two minor things we COULD do:
* suppress echoing the error about the skopeo check for 2.11.0 and the container extraction image + path
* change the usage() method so it's obvious that previous version need not be harcoded anymore

but...

* since we generally WANT to set the previous CSV as a declarative value (set in job-config.json) and 
* falling back to using the last released version is a CI pipeline workaround, not a default behaviour, 
* I actually prefer to see these extra lines of "OH NOES, value passed in is not available, so for the sake of making the Brew/CVP tests pass, let's temporarily override with the older released CSV version" 

Therefore, merging as is.

PR #106, opened 2021-09-21
Change the way we install yarn in .yarn/releases

sleshchenko:
Change the way we install yarn in .yarn/releases

We can't use set-policies anymore since it relies
on an ability to set GitHub token as query param
while header now is required.
See for more details https://github.com/yarnpkg/yarn/issues/7847

Will be tested by build pipeline after merging.


akurinnoy (approved):
LGTM

sleshchenko:
Is triggered in https://main-jenkins-csb-crwqe.apps.ocp4.prod.psi.redhat.com/job/CRW_CI/job/crw-dashboard_2.12/7/

PR #111, opened 2021-09-24
fix: prevent op-metadata sync from needlessly updating createdAt'

mkuznyetsov:
CSV date will be updated, only if there were actual changes after the sync. This will prevent spam of syncing, where only date is being updated


nickboldt:
Also requires:
* eb0dd0bbb (2.x branch)
* 3477f3c84 (2.12 branch)

PR #118, opened 2021-10-01
fix: enable gateway images

tolusha:


flacatus:
LGTM

nickboldt:
cherrypicked to 2.x/2.13 because I'm assuming you also need that fix there.

PR #146, opened 2021-11-19
Using lombok jar assets in brew builds

ghatwala:
signed-off-by:  Amit Ghatwal <ghatwala@us.ibm.com>


codeready-workspaces-plugin-java8/build/build_lombok.sh
```diff
@@ -74,6 +74,7 @@ jarfile="${SCRIPT_DIR}/target/lombok-ls/lombok-${LOMBOK_VERSION}.jar"
 if [[ ! -x ./uploadAssetsToGHRelease.sh ]]; then 
     curl -sSLO "https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/${MIDSTM_BRANCH}/product/uploadAssetsToGHRelease.sh" && chmod +x uploadAssetsToGHRelease.sh
 fi
+./uploadAssetsToGHRelease.sh --delete-assets -v "${CSV_VERSION}" -n ${ASSET_NAME}
```
nickboldt, line 77: we should only delete the release if the LOMBOK_VERSION has changed. Otherwise we're rebuilding and repushing identical bits for no reason. 

See also codeready-workspaces-plugin-java8/get-sources.sh line 45. 

is there a way to control whether or not build_lombok.sh is a no-op if the LOMBOK_VERSION has changed / hasn't changed?

PR #166, opened 2022-01-19
fix: Update Dockerfile for missing slash

ArvinB:
A forward slash is needed after tmp when performing a COPY of multiple files to indicate these are going into a folder.

@nickboldt


nickboldt:
Fixed in sync.sh here: https://github.com/redhat-developer/codeready-workspaces-images/commit/a348cba4403f7393d45f0660bfa024d17b6bf5dd

nickboldt:
and regenerated the dockerfile here: https://github.com/redhat-developer/codeready-workspaces-images/commit/e19398a2cc1e310b3af332d47c929cf5375c367f#diff-7227ac6e7628994256396e0ce853833779b34b3c6b6d7d77bb3d7064466d9f4dR21

PR #170, opened 2022-01-25
FEAT: Update Dockerfile COPY references

ArvinB:
An update to make it easier for downstream projects to easier consume this repository and customize for adding additional plugins.
@nickboldt Can you please take a look?


nickboldt:
Looks good but can you apply this in the upstream repos?

* https://github.com/eclipse-che/che-plugin-registry/
* https://github.com/redhat-developer/codeready-workspaces/tree/crw-2-rhel-8/dependencies/che-plugin-registry/

crw-images repo is MOSTLY generated code from upstream repos, so fixing it in the midstream will just be overwritten the next time we sync downstream

nickboldt (changes requested):
-1, needs to be in the upstream repos.

ArvinB:
Will do, thanks

PR #171, opened 2022-01-25
feat(build): add fake .git folder into build dir

svor:
Signed-off-by: svor <vsvydenk@redhat.com>

Fix for https://issues.redhat.com/browse/CRW-2672

The `eclipse-che/plugin-registry-generator` library should be executed only for git repo. 
To workaround that I propose to add fake .git folder into the builder. 
~~It's not enough just to do make `git init` in the build folder, the repository should already contain some history.~~
It was done by `git init` command

@nickboldt @benoitf  WDYT?


svor:
As we discussed with @benoitf, the real fix should be done in upstream che-plugin-registry in eclipse-che/plugin-registry-generator library.

nickboldt:
If you're tracking the larger/better fix in CRW-2672 (not a GH issue) then I'm happy for this quick workaround to unblock 2.15 and 2.16.

PR #177, opened 2022-02-04
fix: Add devfile-registry Dockerfile transform rule

vinokurig:
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>

* Comment  `COPY ./resources /build/resources` because the `/resources` folder is packed to the tarball and will be unpacked later.
* Remove the rule to add `COPY --from=builder /build/resources /var/www/html/resources` line as, it is already present in the original Dockerfile


nickboldt:
seems reasonable. let's try it!

https://main-jenkins-csb-crwqe.apps.ocp-c1.prod.psi.redhat.com/job/CRW_CI/job/crw-devfileregistry_2.x/797/console

PR #189, opened 2022-02-16
chore: check for python-related directories existence when starting the UDI

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

Fixes failing the container startup when python-related directories don't exist.

### DIFF
--- a/codeready-workspaces-udi/etc/entrypoint.sh
+++ b/codeready-workspaces-udi/etc/entrypoint.sh
@@ -38,13 +38,18 @@ fi
 
 # Setup .venv in the 'venv' volume that should be mounted in HOME/.venv
 mkdir -p "${HOME}"/.venv
+
 if [ ! -f "${HOME}"/.venv/bin/activate ]; then
   echo "${HOME}"/.venv is empty, moving files from "${HOME}"/.venv-tmp/
-  mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
+  if [ -d "${HOME}"/.venv-tmp ]; then
+    mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
+  fi
 fi
 
 # shellcheck source=/dev/null
-source "${HOME}"/.venv/bin/activate
+if [ -f "${HOME}"/.venv/bin/activate ]; then
+  source "${HOME}"/.venv/bin/activate
+fi
 
 if [[ ! -z "${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}" ]]; then
   ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}


nickboldt (changes requested):
> I've checked it works with java, but cannot say anything about python

wouldn't it make sense to try it with a python devfile since this change could impact python ?

since the udi image should have python installed, these shouldn't fail... right? Or are they only created SOMETIMES if the devfile enables python lang server?

vitaliy-guliy:
> > I've checked it works with java, but cannot say anything about python
> 
> wouldn't it make sense to try it with a python devfile since this change could impact python ?

why not? I think the container should be working for all kind of projects.

> since the udi image should have python installed, these shouldn't fail... right? Or are they only created SOMETIMES if the devfile enables python lang server?

`.venv-tmp` dir is created while installing python language server https://github.com/vitaliy-guliy/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-udi/Dockerfile#L128
If the language server ins missing, it should not fail the container.

The error is not related on the devfile. When I try to run the container locally, I have the following output
```
$ docker run -ti quay.io/crw/udi-rhel8:2.16 bash
/home/user/.venv is empty, moving files from /home/user/.venv-tmp/
mv: cannot stat '/home/user/.venv-tmp/*': No such file or directory
```

nickboldt:
The actual error is a docker warning, `warning: Unknown instruction: LN`.

Should be fixed with https://github.com/redhat-developer/codeready-workspaces-images/commit/e96260ff092c7ae6c4704e3202ae9784b8b7797f#diff-488be4ea29f3b560e941ca90c8ec75f32705e5a35e9aa4b3a4545a6ed5f68c03R115

Respinning: https://main-jenkins-csb-crwqe.apps.ocp-c1.prod.psi.redhat.com/job/CRW_CI/job/sync-to-downstream_2.x/3922/

vitaliy-guliy:
> The actual error is a docker warning, `warning: Unknown instruction: LN`.
> 
> Should be fixed with [e96260f#diff-488be4ea29f3b560e941ca90c8ec75f32705e5a35e9aa4b3a4545a6ed5f68c03R115](https://github.com/redhat-developer/codeready-workspaces-images/commit/e96260ff092c7ae6c4704e3202ae9784b8b7797f#diff-488be4ea29f3b560e941ca90c8ec75f32705e5a35e9aa4b3a4545a6ed5f68c03R115)

It's not clear for me how it fixes the issue with python.

nickboldt:
> It's not clear for me how it fixes the issue with python.

Docker doesn't know what `LN` is. Because the previous line was missing a `&& \`. 

My fix should rejoin the lines so the install actually happens

vitaliy-guliy:
> 



> > It's not clear for me how it fixes the issue with python.
> 
> Docker doesn't know what `LN` is. Because the previous line was missing a `&& \`.
> 
> My fix should rejoin the lines so the install actually happens

Wouldn't it be nice to have that check in `entrypoint.sh` as well?

nickboldt:
> Wouldn't it be nice to have that check in entrypoint.sh as well?

maybe, but if we are supposed to have the python venv set up correctly, then entrypoint.sh should FAIL if python isn't configured, not just skip over it.

nickboldt:
After my fix:

```
$➔ docker run -ti quay.io/crw/udi-rhel8:2.16 bash
Unable to find image 'quay.io/crw/udi-rhel8:2.16' locally
2.16: Pulling from crw/udi-rhel8
...
Status: Downloaded newer image for quay.io/crw/udi-rhel8:2.16
/home/user/.venv is empty, moving files from /home/user/.venv-tmp/
(.venv) exit
exit
```

codeready-workspaces-udi/etc/entrypoint.sh
```diff
@@ -38,13 +38,18 @@ fi
 
 # Setup .venv in the 'venv' volume that should be mounted in HOME/.venv
 mkdir -p "${HOME}"/.venv
+
 if [ ! -f "${HOME}"/.venv/bin/activate ]; then
   echo "${HOME}"/.venv is empty, moving files from "${HOME}"/.venv-tmp/
-  mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
+  if [ -d "${HOME}"/.venv-tmp ]; then
+    mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
```
svor, line 45: probably questions to @SDawley:
 
1. "${HOME}"/.venv is empty every time, it was just created in line 40 why do we need 
`if [ ! -f "${HOME}"/.venv/bin/activate ]; then`

2. why do we copy files in entrypoint.sh? 
This operation takes to much time and it makes the workspace starting a bit longer.
Why don't we have it in Dockerfile, then it will happen in build time?
```
mkdir -p "${HOME}"/.venv
mv "${HOME}"/.venv-tmp/* "${HOME}"/.venv
rm -rf "${HOME}"/.venv-tmp
```

vitaliy-guliy: Let me give you an answer.
1. It's not a bid deal if we check for `"${HOME}"/.venv/bin/activate`. `mkdir -p "${HOME}"/.venv` does nothing if the directory is already exists.
2. Copying of files takes 53 seconds when I run the container locally. It's too much.
@nickboldt why not to do everything, that is needed when building the image? It would be nice to add few instructions inside Dockerfile, instead of having them in `entrypoint.sh`.

vitaliy-guliy: I've prepared another one PR with improvement https://github.com/redhat-developer/codeready-workspaces-images/pull/190

vitaliy-guliy:
Persisted by https://github.com/redhat-developer/codeready-workspaces-images/pull/190

PR #218, opened 2022-04-12
fix:cleanup entrypoint.sh for UDI

svor:

It's not possible to execute commands with **sudo** in entrypoint.sh. Can we proceed without removing unused _/etc/alternatives/java_ and _/etc/alternatives/javac_?

Fix for https://issues.redhat.com/browse/CRW-2862

### DIFF
--- a/devspaces-udi/etc/entrypoint.sh
+++ b/devspaces-udi/etc/entrypoint.sh
@@ -53,13 +53,13 @@ fi
 # Use java 8 ONLY if USE_JAVA8 is set to 'true'
 #############################################################################
 if [ "${USE_JAVA8}" == "true" ] && [ ! -z "${JAVA8_HOME}" ]; then
-  export JAVA_HOME=${JAVA8_HOME}
-  export PATH="${JAVA8_HOME}/bin:${PATH}"
+  echo "JAVA_HOME=${JAVA8_HOME}" >> "${HOME}"/.bashrc
+  echo "PATH=${JAVA8_HOME}/bin:${PATH}" >> "${HOME}"/.bashrc
   echo "Java environment set to ${JAVA8_HOME}"
 else
   if [ ! -z "${JAVA11_HOME}" ]; then
-    export JAVA_HOME=${JAVA11_HOME}
-    export PATH="${JAVA11_HOME}/bin:${PATH}"
+    echo "JAVA_HOME=${JAVA11_HOME}" >> "${HOME}"/.bashrc
+    echo "PATH=${JAVA11_HOME}/bin:${PATH}" >> "${HOME}"/.bashrc
     echo "Java environment set to ${JAVA11_HOME}"
   else
     echo "Java environment is not configured"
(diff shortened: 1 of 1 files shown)


nickboldt:
My concern with this change (and the reason I added the alternatives switch) is that in https://issues.redhat.com/browse/CRW-2591 we saw a problem where the debugger and debuggee were not the same Java version.

What do you use in Che's UDI image to support multiple JDKs?

svor:
It doesn't work as expected, I continue investigate the problem

devspaces-udi/etc/entrypoint.sh
```diff
@@ -53,21 +53,13 @@ fi
...
 else
   if [ ! -z "${JAVA11_HOME}" ]; then
-    export JAVA_HOME=${JAVA11_HOME}
-    export PATH="${JAVA11_HOME}/bin:${PATH}"
-    sudo rm -f /etc/alternatives/java
-    sudo ln -s ${JAVA11_HOME}/bin/java /etc/alternatives/java
-    sudo rm -f /etc/alternatives/javac
-    sudo ln -s ${JAVA11_HOME}/bin/javac /etc/alternatives/javac
+    echo "JAVA_HOME=${JAVA11_HOME}" >> "${HOME}"/.bashrc
+    echo "PATH=${JAVA11_HOME}/bin:${PATH}" >> "${HOME}"/.bashrc
```
nickboldt: this should be an `export PATH...` in the .bashrc; same for `export JAVA_HOME=...`

for example in my ~/.bashrc I have:

```
export PATH=$PATH:~/bin
export JAVA_HOME=/usr/lib/jvm/java
export MAVEN_OPTS="-Xms512m -Xmx3000m"
export GITUSER=nickboldt
```

nickboldt (approved):
approved, with my small change. let's run it and see if it breaks java debugger/debuggee support :D

svor:
By making **export** in entrypoint.sh - env was set only into its own environment. By adding ENVs into .bashrc we can solve the problem

nickboldt:
> By making **export** in entrypoint.sh - env was set only into its own environment. By adding ENVs into .bashrc we can solve the problem

Agreed, but should the .bashrc simply define the vars, or EXPORT them to the shell, so they're more pervasive?

svor:
> Agreed, but should the .bashrc simply define the vars, or EXPORT them to the shell, so they're more pervasive?

In case if user will need java8, it could change it by adding environment var in the devfile and then the right value of JAVA_HOME and PATH will be added into .bashrc. 
What do you mean by EXPORT them to the shell?

devspaces-udi/etc/entrypoint.sh
svor: thanks

nickboldt:
> What do you mean by EXPORT them to the shell?

I mean this:

https://github.com/redhat-developer/devspaces-images/pull/218/commits/1e65e2851480d12649270521571d705b2aa82be8

PR #221, opened 2022-04-19
chore: add a mechanism to set java version at runtime

svor:

1. At build time: set JAVA_HOME as **home/user/.java/current** and add it to PATH
2. At runtime: create symlink to java which is needed, by default it's java 11 

Fix for https://issues.redhat.com/browse/CRW-2878


nickboldt:
cpicked to 3.x too.

PR #224, opened 2022-04-20
feat: update docs location

themr0c:
fix https://issues.redhat.com/browse/RHDEVDOCS-3928

Note: I assume the "certificate" link is not present in the dashboard. We have removed this piece of documentation as obsolete.


themr0c:
ping @nickboldt

nickboldt:
Good fix, but wrong file -- need to update the template file and then the json is generated.

So here's your fix + the additional fix for both branches:

* c3a432c94 (3.0)
* 44aad8cc8 (3.x/3.1)

PR #227, opened 2022-04-21
Update sync and get-sources script to get the latest changes for idea-rhel8

vzhukovs:
Update sync and get-sources script to get the latest changes for idea-rhel8.

fixes https://issues.redhat.com/browse/CRW-2736

### DIFF
--- a/devspaces-idea/build/scripts/sync.sh
+++ b/devspaces-idea/build/scripts/sync.sh
@@ -46,29 +46,32 @@ if [[ ! -d "${TARGETDIR}" ]]; then usage; fi
 if [[ "${CSV_VERSION}" == "2.y.0" ]]; then usage; fi
 
 # ignore changes in these files
-echo ".github/
+echo "
 .git/
-.gitignore
-.dockerignore
+.github/
 .idea/
-build/
-devfiles/
+.DS_Store
+*.iml
+make-release.sh
+compatible-ide.json
+devfile.yaml
+README.md
+build/docker/
+build/scripts/sync.sh
 doc/
 kubernetes/
-*.iml
-ide-packaging
-projector-server-assembly
-static-assembly
-/container.yaml
-/content_sets.*
-/cvp.yml
-/cvp-owners.yml
-/patches/README.md
-README.md
+devfiles/
+container.yaml
+content_sets.*
+cvp.yml
+cvp-owners.yml
 get-source*.sh
 tests/basic-test.yaml
 sources
-make-release.sh
+asset-ide-packaging.tar.gz
+asset-che-plugin-assembly.zip
+asset-projector-server-assembly.zip
+asset-static-assembly.tar.gz
 " > /tmp/rsync-excludes
 echo "Rsync ${SOURCEDIR} to ${TARGETDIR}"
 rsync -azrlt --checksum --exclude-from /tmp/rsync-excludes --delete "${SOURCEDIR}"/ "${TARGETDIR}"/
@@ -89,8 +92,6 @@ sed_in_place() {
 sed_in_place -r \
   `# Remove registry so build works in Brew` \
   -e "s#FROM (registry.access.redhat.com|registry.redhat.io)/#FROM #g" \
-  `# Remove unused Python packages (support for PyCharm not included in CRW)` \
-  -e "/# Python support/d" -e "/python2 python39 \\\\/d" \
   "${TARGETDIR}"/Dockerfile
 
 # Overwrite default configuration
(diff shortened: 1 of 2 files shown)


devspaces-idea/build/scripts/sync.sh
```diff
@@ -46,29 +46,32 @@ if [[ ! -d "${TARGETDIR}" ]]; then usage; fi
...
 *.iml
-ide-packaging
-projector-server-assembly
-static-assembly
-/container.yaml
-/content_sets.*
-/cvp.yml
-/cvp-owners.yml
-/patches/README.md
+make-release.sh
```
nickboldt, line 55: if you're going to reorder everything here, why not make it alphabetical?

devspaces-idea/build/scripts/sync.sh
```diff
@@ -46,29 +46,32 @@ if [[ ! -d "${TARGETDIR}" ]]; then usage; fi
...
+compatible-ide.json
+devfile.yaml
 README.md
+build/docker/
+build/scripts/sync.sh
```
nickboldt, line 60: exclude only these two items in build/ and copy/sync the rest?

devspaces-idea/build/scripts/sync.sh
```diff
@@ -46,29 +46,32 @@ if [[ ! -d "${TARGETDIR}" ]]; then usage; fi
...
+devfiles/
+container.yaml
+content_sets.*
+cvp.yml
+cvp-owners.yml
 get-source*.sh
 tests/basic-test.yaml
 sources
-make-release.sh
+asset-ide-packaging.tar.gz
```
nickboldt, line 71: do these asset files exist in upstream? if not, they'll never be synced to mid or downstream; afair they're CREATED with build.sh or get-sources.sh so do they need to be listed here?

devspaces-idea/build/scripts/sync.sh
```diff
@@ -46,29 +46,32 @@ if [[ ! -d "${TARGETDIR}" ]]; then usage; fi
...
-.dockerignore
+.github/
 .idea/
-build/
-devfiles/
-doc/
-kubernetes/
+.DS_Store
 *.iml
-ide-packaging
```
nickboldt, line 59: so `ide-packaging` and `projector-server-assembly` and `static-assembly` folders no longer exist, replaced by zips/tars?

devspaces-idea/get-sources.sh
```diff
@@ -10,7 +10,7 @@ forceBuild=0
 # so that rhpkg build is simply a brew wrapper (using get-sources.sh -f)
 PULL_ASSETS=0
 
-idePackagingUrl=https://download-cdn.jetbrains.com/idea/ideaIC-2020.3.3.tar.gz
+idePackagingUrl=https://download-cdn.jetbrains.com/idea/ideaIC-2020.3.4.tar.gz
```
nickboldt, line 13: too bad we can't move up to 2022 version :(

nickboldt (approved):
+0, probably ok but with questions.

devspaces-idea/get-sources.sh
vzhukovs: 2022 hasn't supported by Projector yet. But it is possible to update to 2021, but this will be the next step. Need to make a lot of changes upstream first and test it.

devspaces-idea/build/scripts/sync.sh
vzhukovs: There was files, not folders. But yes, assets names have been updated to meaningful ones.

devspaces-idea/build/scripts/sync.sh
vzhukovs: >do these asset files exist in upstream?

No, they don't exist in upstream, they are creating during the build time. Should these assets be removed from the exclude list?

vzhukovs:
@nickboldt updated PR.

devspaces-idea/build/scripts/sync.sh
```diff
@@ -89,8 +84,6 @@ sed_in_place() {
 sed_in_place -r \
   `# Remove registry so build works in Brew` \
   -e "s#FROM (registry.access.redhat.com|registry.redhat.io)/#FROM #g" \
-  `# Remove unused Python packages (support for PyCharm not included in CRW)` \
```
nickboldt, line 92: why remove this?

vzhukovs: We use utility `yq` to operate with Devfile. `yq` requires python packages.

nickboldt: are you prefetching yq to install it offline? 

I only see you mention yq in https://github.com/che-incubator/jetbrains-editor-images/blob/main/static/ide-projector-launcher.sh#L115-L122 and it seems to be assumed to be already installed. 

Is yq now bundled with python 3.9?

https://github.com/che-incubator/jetbrains-editor-images/blob/main/Dockerfile#L52

is yq and its python deps installed via one of these tarballs?

https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-idea/Dockerfile#L64-L80

vzhukovs: Yes, you were right. Just rechecked. `yq` expects to be installed in UDI image to operate with Devfile. For local run it is not used due to non-existent Devfile in container. So, reverted `sed` parameter.

nickboldt:
do we need this cherrypicked to the 3.0 branch too?

azatsarynnyy:
I've updated the `Fix Version` to 3.1

PR #232, opened 2022-04-22
refactor CRW_ prefix in openshift templates and operator code to use DS_ prefix (CRW-2822)

SDawley:
Updating variables from CRW etc to DS

https://issues.redhat.com/browse/CRW-2822


SDawley:
@nickboldt I think I got most of them but I left CRW's that were referring to previous versions (like notes that say when something was added) or references to JIRAs.

nickboldt (approved):
+1, with the caveat that the new DS_CI folder doesn't yet exist... so readmes will be wrong until CRW-2839 is solved.

PR #239, opened 2022-05-02
Provide fixed packages for idea-rhel8

vzhukovs:


nickboldt:
I don't love the filename cve-fixed-packages since one of the two packages listed is just "latest" and has no hardcoded value for a cve fix; also there's no file extension. 

could we change that to "required-rpms.txt" or something?

Or... instead of depending on this extra file downstream, could you just:

`sed -r -i Dockerfile "s#\$\(cat /tmp/cve-fixed-packages)#libsecret libsecret-devel#"`

PR #254, opened 2022-06-03
fix: remove meta.yaml if devfile.yaml.${arch} exists

svor:
I am not sure if changes from https://github.com/redhat-developer/devspaces/pull/739 will be synced automatically.

What issues does this PR fix or reference?
https://issues.redhat.com/browse/CRW-3034


svor:
changes were synced

PR #261, opened 2022-06-24
feat: Productize che-code as an OpenShift Dev Spaces editor (CRW-2744)

RomanNikitenko:
Productize che-code as an OpenShift Dev Spaces editor

See [CRW-2744](https://issues.redhat.com/browse/CRW-2744)

### DIFF
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -22,7 +22,7 @@ RUN cat /mnt/rootfs/etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}:
 
 COPY asset-*.tar.gz /tmp/assets/
 
-RUN tar -xf /tmp/assets/*.tar.gz -C /mnt/rootfs && \
+RUN for f in /tmp/assets/*.tar.gz; do tar -xf "$f" -C /mnt/rootfs; done && \
     mv /mnt/rootfs/che-machine-exec /mnt/rootfs/bin/machine-exec
 
 RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" "/mnt/rootfs/etc/group" "/mnt/rootfs/projects" "/mnt/rootfs/entrypoint*.sh" "/mnt/rootfs/checode-linux-libc" ; do\
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -11,7 +11,7 @@
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8
 FROM ubi8:8.6-754.1655117782 as ubi-builder
 
-RUN mkdir -p /mnt/rootfs /mnt/rootfs/projects /mnt/rootfs/home/che
+RUN mkdir -p /mnt/rootfs/projects /mnt/rootfs/home/che
 RUN yum install --installroot /mnt/rootfs tar gzip brotli libstdc++ coreutils glibc-minimal-langpack --releasever 8 --setopt install_weak_deps=false --nodocs -y && yum --installroot /mnt/rootfs clean all
 RUN rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*
 
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -20,22 +20,18 @@ WORKDIR /mnt/rootfs
 RUN mkdir -p /mnt/rootfs/projects && mkdir -p /mnt/rootfs/home/che && mkdir -p /mnt/rootfs/bin/
 RUN cat /mnt/rootfs/etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > /mnt/rootfs/home/che/.passwd.template \
     && cat /mnt/rootfs/etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > /mnt/rootfs/home/che/.group.template
-RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" "/mnt/rootfs/etc/group" "/mnt/rootfs/projects" ; do\
+
+RUN tar -xf /tmp/assets/asset-machine-exec.tar.gz && \
+    mv che-machine-exec /mnt/rootfs/bin/machine-exec
+RUN tar -xf /tmp/assets/asset-linux-libc.tar.gz -C /mnt/rootfs
+RUN tar -xf /tmp/assets/asset-entrypoints.tar.gz -C /mnt/rootfs
+
+RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" "/mnt/rootfs/etc/group" "/mnt/rootfs/projects" "/mnt/rootfs/entrypoint*.sh" "/mnt/rootfs/checode-linux-libc" ; do\
            chgrp -R 0 ${f} && \
            chmod -R g+rwX ${f}; \
        done
 
 COPY asset-*.tar.gz /tmp/assets/
-
-RUN tar -xf /tmp/assets/asset-machine-exec.tar.gz && \
-    mv che-machine-exec /mnt/rootfs/bin/machine-exec && \
-    chown 0:0 /mnt/rootfs/bin/machine-exec
-
-RUN tar -xf /tmp/assets/asset-linux-libc.tar.gz -C /mnt/rootfs && \
-    chown 0:0 /mnt/rootfs/checode-linux-libc
-
-RUN tar -xf /tmp/assets/asset-entrypoints.tar.gz -C /mnt/rootfs && \
-    chmod 755 /mnt/rootfs/entrypoint*.sh
 
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
 FROM ubi8-minimal:8.6-751.1655117800
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -9,7 +9,7 @@
 #   Red Hat, Inc. - initial API and implementation
 
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8
-FROM ubi8:8.6-754.1655117782 as ubi-builder
+FROM ubi8:8.6-855 as ubi-builder
 
 RUN mkdir -p /mnt/rootfs/projects /mnt/rootfs/home/che
 RUN yum install --installroot /mnt/rootfs tar gzip brotli libstdc++ coreutils glibc-minimal-langpack --releasever 8 --setopt install_weak_deps=false --nodocs -y && yum --installroot /mnt/rootfs clean all
@@ -31,7 +31,7 @@ RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" 
        done
 
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
-FROM ubi8-minimal:8.6-751.1655117800
+FROM ubi8-minimal:8.6-854
 COPY --from=ubi-builder /mnt/rootfs/ /
 ENV HOME=/home/che
 USER 1001
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -1,0 +1,60 @@
+# Copyright (c) 2022 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+FROM registry.access.redhat.com/ubi8:8.6-754 as ubi-builder
+
+RUN mkdir -p /mnt/rootfs
+RUN yum install --installroot /mnt/rootfs tar gzip brotli libstdc++ coreutils glibc-minimal-langpack --releasever 8 --setopt install_weak_deps=false --nodocs -y && yum --installroot /mnt/rootfs clean all
+RUN rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*
+
+WORKDIR /mnt/rootfs
+
+RUN mkdir -p /mnt/rootfs/projects && mkdir -p /mnt/rootfs/home/che && mkdir -p /mnt/rootfs/bin/
+RUN cat /mnt/rootfs/etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > /mnt/rootfs/home/che/.passwd.template \
+    && cat /mnt/rootfs/etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > /mnt/rootfs/home/che/.group.template
+RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" "/mnt/rootfs/etc/group" "/mnt/rootfs/projects" ; do\
+           chgrp -R 0 ${f} && \
+           chmod -R g+rwX ${f}; \
+       done
+
+COPY asset-*.tar.gz /tmp/assets/
+
+RUN tar -xf /tmp/assets/asset-machine-exec.tar.gz && \
+    mv che-machine-exec /mnt/rootfs/bin/machine-exec && \
+    chown 0:0 /mnt/rootfs/bin/machine-exec
+
+RUN tar -xf /tmp/assets/asset-linux-libc.tar.gz -C /mnt/rootfs && \
+    chown 0:0 /mnt/rootfs/checode-linux-libc
+
+RUN tar -xf /tmp/assets/asset-entrypoints.tar.gz -C /mnt/rootfs && \
+    chmod 755 /mnt/rootfs/entrypoint*.sh
+
+FROM registry.access.redhat.com/ubi8-minimal:8.6-751
+COPY --from=ubi-builder /mnt/rootfs/ /
+ENV HOME=/home/che
+USER 1001
+ENTRYPOINT /entrypoint.sh
+
+ENV SUMMARY="Red Hat OpenShift Dev Spaces code container" \
+    DESCRIPTION="Red Hat OpenShift Dev Spaces code container" \
+    PRODNAME="devspaces" \
+    COMPNAME="code-rhel8"
+LABEL summary="$SUMMARY" \
+      description="$DESCRIPTION" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="$DESCRIPTION" \
+      io.openshift.tags="$PRODNAME,$COMPNAME" \
+      com.redhat.component="$PRODNAME-$COMPNAME-container" \
+      name="$PRODNAME/$COMPNAME" \
+      version="3.1" \
+      license="EPLv2" \
+      maintainer="Roman Nikitenko <rnikiten@redhat.com>" \
+      io.openshift.expose-services="" \
+      usage=""
--- a/devspaces-code/Dockerfile
+++ b/devspaces-code/Dockerfile
@@ -11,13 +11,12 @@
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8
 FROM ubi8:8.6-754.1655117782 as ubi-builder
 
-RUN mkdir -p /mnt/rootfs
+RUN mkdir -p /mnt/rootfs /mnt/rootfs/projects /mnt/rootfs/home/che
 RUN yum install --installroot /mnt/rootfs tar gzip brotli libstdc++ coreutils glibc-minimal-langpack --releasever 8 --setopt install_weak_deps=false --nodocs -y && yum --installroot /mnt/rootfs clean all
 RUN rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*
 
 WORKDIR /mnt/rootfs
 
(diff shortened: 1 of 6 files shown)


devspaces-code/build/scripts/collect-assets.sh
```diff
@@ -0,0 +1,45 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+LINUX_LIBC_IMAGE=linux-libc-amd64:latest
+MACHINE_EXEC_IMAGE=quay.io/eclipse/che-machine-exec:7.39.1
```
RomanNikitenko, line 14: it makes sense to get machine exec version from upstream

devspaces-code/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+FROM registry.access.redhat.com/ubi8:8.6-754 as ubi-builder
+
+RUN mkdir -p /mnt/rootfs
+RUN yum install --installroot /mnt/rootfs tar gzip brotli libstdc++ coreutils glibc-minimal-langpack --releasever 8 --setopt install_weak_deps=false --nodocs -y && yum --installroot /mnt/rootfs clean all
+RUN rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/dnf* /mnt/rootfs/var/log/yum.*
+
+WORKDIR /mnt/rootfs
+
+RUN mkdir -p /mnt/rootfs/projects && mkdir -p /mnt/rootfs/home/che && mkdir -p /mnt/rootfs/bin/
```
vzhukovs, line 19: `/mnt/rootfs/bin/` directory should be existed at this moment, need to check if it's true, then this command is redundant.

devspaces-code/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN cat /mnt/rootfs/etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > /mnt/rootfs/home/che/.passwd.template \
+    && cat /mnt/rootfs/etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > /mnt/rootfs/home/che/.group.template
+RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" "/mnt/rootfs/etc/group" "/mnt/rootfs/projects" ; do\
+           chgrp -R 0 ${f} && \
+           chmod -R g+rwX ${f}; \
+       done
+
+COPY asset-*.tar.gz /tmp/assets/
```
vzhukovs, line 27: Line 27-37 could be safely moved to the line 22 and in for loop there `/mnt/rootfs/bin/machine-exec`, `/mnt/rootfs/checode-linux-libc`, `/mnt/rootfs/entrypoint*.sh` could be added. It's just an optimisation, but up to you.

devspaces-code/build/scripts/sync.sh
```diff
@@ -0,0 +1,91 @@
...
+  elif [ "${SHORT_UNAME:0:5}" == "Linux" ]; then
+    sed -i "$@"
+  fi
+}
+
+sed_in_place -r \
+  `# Update ubi8 image name` \
+  -e "s#ubi8/ubi:#ubi8:#g" \
+  `# Remove registry so build works in Brew` \
+  -e "s#FROM (registry.access.redhat.com|registry.redhat.io)/#FROM #g" \
```
vzhukovs, line 82: As far as you have stored Dockerfile in the current repository, you can omit this two sed's and provide correct image names directly in Dockerfile

vzhukovs: Because after syncing upstream with midstream, you'll have adjusted and committed Dockerfile in the repository.

devspaces-code/Dockerfile
nickboldt: you can also do a single command:
```
RUN mkdir -p /mnt/rootfs/projects /mnt/rootfs/home/che /mnt/rootfs/bin/
```

devspaces-code/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN tar -xf /tmp/assets/asset-entrypoints.tar.gz -C /mnt/rootfs && \
+    chmod 755 /mnt/rootfs/entrypoint*.sh
+
+FROM registry.access.redhat.com/ubi8-minimal:8.6-751
+COPY --from=ubi-builder /mnt/rootfs/ /
+ENV HOME=/home/che
+USER 1001
+ENTRYPOINT /entrypoint.sh
+
+ENV SUMMARY="Red Hat OpenShift Dev Spaces code container" \
```
nickboldt, line 45: are we calling this "code" or "vscode" or "Visual Studio Code" ?

devspaces-code/build/scripts/collect-assets.sh
nickboldt: you also can't use che-machine-exec in downstream. Use the productized version from the operator's CSV 

https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-operator-bundle/manifests/devspaces.csv.yaml#L1119-L1122

devspaces-code/build/scripts/collect-assets.sh
```diff
@@ -0,0 +1,45 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+LINUX_LIBC_IMAGE=linux-libc-amd64:latest
```
nickboldt, line 13: this smells like a single-arch implementation. Please make this multiarch or we can't use this in downstream.

devspaces-code/build/scripts/sync.sh
nickboldt: if you omit this sed and instead just make the change to the dockefile in the repo, you will have a dockerfile that can ONLY be run in Brew (not used locally).

So... to make midstream development easier, we transform the midstream with `FROM registry.access.redhat.co/foo` to `FROM foo` in downsteam and save that change ONLY to the brew version of the dockerfile in pkgs.devel

devspaces-code/get-sources.sh
```diff
@@ -0,0 +1,96 @@
...
+if [[ ${PULL_ASSETS} -eq 1 ]]; then
+  ./build/scripts/collect-assets.sh 
+
+  if [[ ! -f "asset-linux-libc.tar.gz" ]]; then
+    log "[ERROR] 'asset-linux-libc.tar.gz' not found, so nothing to build."
+    exit 1;
+  fi
+
+  if [[ ! -f "asset-machine-exec.tar.gz" ]]; then
+    log "[ERROR] 'asset-machine-exec.tar.gz' not found, so nothing to build."
```
nickboldt, line 42: maybe a better errror message is "not found, so cannot proceed with build" ?

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: sorry, could you clarify what's the problem here?
I don't pull the image, I prepare `linux-libc-amd64:latest` image and put the corresponding content as asset, so I don't use the image within downstream...

devspaces-code/build/scripts/collect-assets.sh
vzhukovs: >you also can't use che-machine-exec in downstream

Why not? It's just a copied binary from the upstream image and placed as asset.

vzhukovs: Or do you mean, to use this image `registry-proxy.engineering.redhat.com/rh-osbs/devspaces-machineexec-rhel8:3.1` to fetch the asset instead of `quay.io/eclipse/che-machine-exec:7.39.1` ?

devspaces-code/build/scripts/collect-assets.sh
```diff
@@ -0,0 +1,45 @@
...
+SCRIPT_PATH="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
+BASE_DIR_PATH="${SCRIPT_PATH}/../.."
+DOCKERFILES_PATH="${BASE_DIR_PATH}/build/dockerfiles"
+
+#### linux-libc-content ####
+collect_linux_libc_content_assets() {
+    docker build -f "${DOCKERFILES_PATH}/linux-libc.Dockerfile" -t $LINUX_LIBC_IMAGE .
+
+    id="$(docker create $LINUX_LIBC_IMAGE)"
+    docker cp "$id":/checode-linux-libc - | gzip -9 > asset-linux-libc.tar.gz
```
nickboldt, line 25: if this is expected to be different by arch, then you need an arch-specific filename.

devspaces-code/build/scripts/collect-assets.sh
nickboldt: if the resulting docker container (and then tarball) is expected to be different by arch, then you need an arch-specific filename for each asset file, so that:
* each new file doesn't collide when pushing into pkgs.devel
* only the tarball that's compatible with s390x is used on s390x brew build (and same for ppc64le, amd64/x86_64)

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: > you also can't use che-machine-exec in downstream

We don't use `collect-assets.sh` in downstream
It's used to get assets, so when `get-sources.sh` works

devspaces-code/build/scripts/sync.sh
vzhukovs: The Dockerfile is not fetched from the upstream, it committed to the midstream directly. Calling sync.sh will commit changes at first time and then this sed instruction will never be reached, am I right?

devspaces-code/build/scripts/collect-assets.sh
vzhukovs: Got it, then image from https://catalog.redhat.com/software/containers/devspaces/machineexec-rhel8/622bb09aae289285d6fcda48 should be used as image from which `machine-exec` binary is copied.

I'll update `devspaces-idea` later, because it also uses `machine-exec` binary under the hood.

devspaces-code/get-sources.sh
RomanNikitenko: applied, thank you!

devspaces-code/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    mv che-machine-exec /mnt/rootfs/bin/machine-exec && \
+    chown 0:0 /mnt/rootfs/bin/machine-exec
+
+RUN tar -xf /tmp/assets/asset-linux-libc.tar.gz -C /mnt/rootfs && \
+    chown 0:0 /mnt/rootfs/checode-linux-libc
+
+RUN tar -xf /tmp/assets/asset-entrypoints.tar.gz -C /mnt/rootfs && \
+    chmod 755 /mnt/rootfs/entrypoint*.sh
+
+FROM registry.access.redhat.com/ubi8-minimal:8.6-751
```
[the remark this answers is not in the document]
RomanNikitenko: done, thank you!

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: As far as I understand we should apply [this way](https://issues.redhat.com/browse/CRW-2744?focusedCommentId=20517790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20517790) for machine-exec

@nickboldt @vzhukovs

devspaces-code/build/scripts/sync.sh
RomanNikitenko: @vzhukovs 
I checked that part - you are right, so I've removed those lines...
thank you!

devspaces-code/build/scripts/collect-assets.sh
nickboldt: we can't use [fetch-artifacts-url](https://osbs.readthedocs.io/en/latest/users.html#fetch-artifacts-url-yaml) (see [CRW-3080](https://issues.redhat.com/browse/CRW-3080)) or cachito ([CRW-2444](https://issues.redhat.com/browse/CRW-2444)) for an asset inside a container we build. 

Is there any way we can just have the vscode editor mount a second container w/ machineexec-rhel8 in it? Theia does that ... why can't vscode do similar?

devspaces-code/Dockerfile
nickboldt: I've opened https://issues.redhat.com/browse/CRW-3098 to make sure we do  this correctly. Also included ref to idea container, in case `"Red Hat OpenShift Dev Spaces idea container"` should be changed to something more product-license-compliant like `"Red Hat OpenShift Dev Spaces - JetBrains IntelliJ IDEA IDE container"`

devspaces-code/Dockerfile
RomanNikitenko: done, thanks!

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: > Would it be possible to pull from quay.io/devspaces/machineexec-rhel8:3.1 instead? then you're at least aligned to the current release, not the previous GA.

I think yes, we could use current release version of the `machine-exec`.
At the same time it means that `devspaces/code-rhel8:3.1` depends on `devspaces/machineexec-rhel8:3.1`.

Is it OK for the productization process?
It means that we should run jobs related to the `machine-exec`, wait for the corresponding images, and only then run jobs related to the `code` and `idea`(`idea` uses `machine-exec` as well)...

nickboldt: It's not ideal to have cross-image dependencies but we already have that with 
* theia-dev -> theia -> theia-endpoint
... so why not also have 
* machineexec -> [idea and vscode] ? 

Given how infrequently machineexec is updated, this is probably not too painful.

devspaces-code/Dockerfile
```diff
@@ -11,13 +11,12 @@
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8
 FROM ubi8:8.6-754.1655117782 as ubi-builder
 
-RUN mkdir -p /mnt/rootfs
+RUN mkdir -p /mnt/rootfs /mnt/rootfs/projects /mnt/rootfs/home/che
```
nickboldt, line 14: you forgot `/mnt/rootfs/bin/`

Why not just do this?

```
RUN mkdir -p /mnt/rootfs/bin/ /mnt/rootfs/projects/ /mnt/rootfs/home/che/

```

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: OK, I see, thanks for the response!

devspaces-code/Dockerfile
RomanNikitenko: I didn't forget it, please see https://github.com/redhat-developer/devspaces-images/pull/261#discussion_r906067250

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: I updated the corresponding part, please take a look when you have a chance...

devspaces-code/build/scripts/collect-assets.sh
```diff
@@ -10,8 +10,10 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
+# the version is updated automatically (see sync.sh)
+DS_VERSION=3.1
```
nickboldt, line 14: I don't think it's wise to hard code this here, even with your sed transform below. 

Why not pass the version into collect-assets.sh as `./build/scripts/collect-assets.sh -t ${DS_VERSION}` in line 34 of get-sources.sh https://github.com/redhat-developer/devspaces-images/pull/261/files#diff-42c11321148052a59ae01c6de12527a8418f07e86557c238c71c6ade40d83102R34 ? 

Or for that matter... you don't need a separate collect-assets file. You could fold that script into get-sources.sh and not have to call a separate script. Simpler?

devspaces-code/build/scripts/sync.sh
```diff
@@ -76,10 +76,15 @@ sed_in_place() {
...
 sed_in_place -r \
-  `# Update DevSpaces version` \
+  `# Update DevSpaces version for Dockerfile` \
   -e "s/version=.*/version=\"$DS_VERSION\" \\\/" \
   "${TARGETDIR}"/Dockerfile
 
+sed_in_place -r \
+  `# Update DevSpaces version for collect-assets script` \
+  -e "s/DS_VERSION=.*/DS_VERSION=$DS_VERSION/" \
+  "${TARGETDIR}"/build/scripts/collect-assets.sh
```
nickboldt, line 86: this is needlessly complex. Just fold the contents of collect-assets into get-sources.sh and you'll have access to the correct `DS_VERSION` variable so you can update the Dockerfile to include the correct version.

devspaces-code/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+WORKDIR /mnt/rootfs
+
+RUN cat /mnt/rootfs/etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > /mnt/rootfs/home/che/.passwd.template \
+    && cat /mnt/rootfs/etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > /mnt/rootfs/home/che/.group.template
+
+COPY asset-*.tar.gz /tmp/assets/
+
+RUN tar -xf /tmp/assets/*.tar.gz -C /mnt/rootfs && \
+    mv /mnt/rootfs/che-machine-exec /mnt/rootfs/bin/machine-exec
```
nickboldt, line 26: this could fail if `/mnt/rootfs/bin/` doesn't already exist.

devspaces-code/Dockerfile
nickboldt: do you plan to make this change in Dockerfile, or wait until 3098 has more clarity?

devspaces-code/build/scripts/collect-assets.sh
nickboldt: since rpms installed vary by arch, I would strongly suggest having the arch in the asset-*.tgz filename, unless you know for sure that the resulting tarball is the same on all three arches.

devspaces-code/Dockerfile
RomanNikitenko: image

why do you think that 
> /mnt/rootfs/bin/ doesn't already exist

?

devspaces-code/build/scripts/sync.sh
```diff
@@ -76,10 +76,15 @@ sed_in_place() {
...
+
 cat << EOT >> "${TARGETDIR}"/build/dockerfiles/linux-libc.Dockerfile
```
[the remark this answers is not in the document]
RomanNikitenko: in **the midstream** I do something like

image

to create a container and get some content as asset.

In **the upstream** we don't create a container using that image, so it's the redundant change for the upstream...

That change allows me to avoid an error like `Error response from daemon: No command specified`.
If you know the better solution - please let me know - I would really appreciate it...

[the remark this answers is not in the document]
nickboldt: Please don't use docker; use podman. 

If the container image in midstream isn't being used then why would adding a `CMD` line break anything upstream?

devspaces-code/Dockerfile
nickboldt: I don't know, I'm just saying you could be explicit about creating it :)

RomanNikitenko: There is a comment from another reviewer that it's redundant.
Also I checked that the folder exists, so I don't see a problem here...

devspaces-code/Dockerfile
RomanNikitenko: I don't see what I can change here without clarification from 3098 or another source.

devspaces-code/build/scripts/sync.sh
RomanNikitenko: Do you mean that I can rely that `get-sources.sh` gets `DS_VERSION`(`CSV_VERSION`)?

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: please see https://github.com/redhat-developer/devspaces-images/pull/261#discussion_r911940270

devspaces-code/build/scripts/sync.sh
RomanNikitenko: I provided the corresponding changes based on the statement we can rely that `get-sources.sh` gets `CSV_VERSION`

devspaces-code/build/scripts/sync.sh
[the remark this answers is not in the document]
RomanNikitenko: I tried to use podman instead of docker, but got an error:

image

Also I see that docker is used in another place of the `devspaces-images` repo, so maybe it's not critical right now and we could come back to this problem later.
wdyt?

devspaces-code/build/scripts/collect-assets.sh
nickboldt: if `docker cp` fails in Jenkins (as I fear it might) you can set up a volume mount and then tarball that volume mount:

Examples:
* https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/build/build_python.sh#L61-L80
* https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/build/build_golang.sh#L63-L90

RomanNikitenko: I wonder how it works for `devspaces-idea` [here](https://github.com/redhat-developer/devspaces-images/blob/8e608d5b2a632dc65206bb2777213deff9e09679/devspaces-idea/projector.sh#L223), for example, if there is no docker

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: added `${ARCH}` part to assets

nickboldt:
Merged: rebased and moved new code.jenkins into DS_CI folder for 3.1/3.x: https://gitlab.cee.redhat.com/codeready-workspaces/crw-jenkins/-/merge_requests/1186

devspaces-code/build/scripts/collect-assets.sh
RomanNikitenko: thanks!

PR #262, opened 2022-06-25
Switch to .NET 6 (add support for s390x; remove dotnet 3.1)

omajid:
We could also keep .NET Core 3.1 (though it will go EOL later this year - 2022).

.NET 6 brings several advantages, incluing support for additional architectures such as ARM 64 (aarch64) and IBM Z (s390x).

https://issues.redhat.com/browse/CRW-2469


devspaces-devfileregistry/arbitrary-users-patch/base_images
```diff
@@ -1,5 +1,6 @@
 che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7
 che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
+che-dotnet-6.0          registry.access.redhat.com/ubi8/dotnet-6.0
```
omajid, line 3: Not sure what this file is for, because it doesn't contain 3.1 either.

devspaces-udi/Dockerfile
```diff
@@ -26,7 +26,7 @@ ENV \
     E2FSPROGS_VERSION="1.46.5" \
     LD_LIBRARY_PATH="/usr/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}" \
     CPATH="/usr/include${CPATH:+:${CPATH}}" \
-    DOTNET_RPM_VERSION=3.1 \
+    DOTNET_RPM_VERSIONS="6.0 3.1" \
```
omajid, line 29: Are there any consumers of this env var? The rename might break them if they are looking for it.

devspaces-udi/Dockerfile
```diff
@@ -257,7 +260,7 @@ RUN \
     echo "========" && \
     echo -n "clangd:    "; clangd --version; \
     if [[ -f /usr/bin/dotnet ]]; then \
-      echo -n "dotnet:    "; dotnet --version; \
+      echo -n "dotnet:    "; dotnet --info; \
```
omajid, line 263: `dotnet --info` will show more details, including all the different .NET versions (and runtimes) that are installed. On my Fedora machine, the output looks like this:

```
$ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.105
 Commit:    1c35735293

Runtime Environment:
 OS Name:     fedora
 OS Version:  36
 OS Platform: Linux
 RID:         fedora.36-x64
 Base Path:   /usr/lib64/dotnet/sdk/6.0.105/

Host (useful for support):
  Version: 6.0.5
  Commit:  70ae3df4a6

.NET SDKs installed:
  3.1.419 [/usr/lib64/dotnet/sdk]
  6.0.105 [/usr/lib64/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.25 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.5 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.25 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.5 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```

Is that too verbose?

devspaces-udi/Dockerfile
```diff
@@ -164,11 +164,14 @@ RUN \
     echo "clangd -pretty" > ${HOME}/che/ls-clangd/launch.sh && \
     chmod +x ${HOME}/che/ls-clangd/launch.sh
     ########################################################################
-    # Dotnet (x64 only) (Tech Preview)
+    # .NET (dotnet) (aarch64, s390x and x64 only) (Tech Preview)
     ########################################################################
 RUN \
     if [[ "$(uname -m)" == 'x86_64' ]]; then \
-        dnf -y -q --setopt=tsflags=nodocs install dotnet dotnet-host dotnet-hostfxr-${DOTNET_RPM_VERSION} dotnet-runtime-${DOTNET_RPM_VERSION} dotnet-sdk-${DOTNET_RPM_VERSION}; \
+        dnf -y -q --setopt=tsflags=nodocs install dotnet-sdk-3.1 \
```
omajid, line 171: Installing the SDK is sufficient to install all the dependencies, including the host, hostfxr, runtime, etc.

It does install an extra 3.1 SDK, though, in addition to the 6.0 SDK (installed a few lines below). If we wanted to reduce the container image size, we could install just the `dotnet-runtime-3.1` and `aspnetcore-runtime-3.1`. The 6.0 SDK is generally good enough to build 3.1 projects, but there may be backward compatibility issues that some users work around by explicitly using the 3.1 SDK.

omajid:
I dont have a way to test this; building the container fails at:


```
STEP 7/25: COPY etc/docker.sh /usr/local/bin/docker                                                                                                                                 
--> 07f999693ca                                                                                                                                                                     
STEP 8/25: COPY lombok-${LOMBOK_VERSION}.jar /lombok.jar                                                                                                                            
Error: error building at STEP "COPY lombok-${LOMBOK_VERSION}.jar /lombok.jar": checking on sources under "/home/omajid/devel/devspaces-images/devspaces-udi": copier: stat: "/lombok-1.18.22.jar": no such file or directory
```

devspaces-devfileregistry/arbitrary-users-patch/base_images
nickboldt: this is a carryover from upstream; we don't use it in devspaces. It should probably be removed from the files we sync from upstream -> downstream.

devspaces-udi/Dockerfile
nickboldt: it's definitely verbose but I like that it dumps both 3.1 and 6.0.

devspaces-udi/Dockerfile
nickboldt: do we really need BOTH 3.1 and 6.0?

devspaces-udi/Dockerfile
```diff
@@ -164,11 +164,14 @@ RUN \
     echo "clangd -pretty" > ${HOME}/che/ls-clangd/launch.sh && \
     chmod +x ${HOME}/che/ls-clangd/launch.sh
     ########################################################################
-    # Dotnet (x64 only) (Tech Preview)
+    # .NET (dotnet) (aarch64, s390x and x64 only) (Tech Preview)
```
nickboldt, line 167: aarch is not a supported arch for Dev Spaces, so no need to mention it here or trap for it on line 173.

devspaces-operator/pkg/deploy/image-puller/imagepuller_test.go
```diff
@@ -438,6 +438,7 @@ func TestEnvVars(t *testing.T) {
 				"RELATED_IMAGE_che_cpp_rhel7_devfile_registry_image_G4XDGNR":                       "quay.io/eclipse/che-cpp-rhel7",
 				"RELATED_IMAGE_che_dotnet_2_2_devfile_registry_image_G4XDGNR":                      "quay.io/eclipse/che-dotnet-2.2",
 				"RELATED_IMAGE_che_dotnet_3_1_devfile_registry_image_G4XDGNR":                      "quay.io/eclipse/che-dotnet-3.1",
+				"RELATED_IMAGE_che_dotnet_6_0_devfile_registry_image_G4XDGNR":                      "quay.io/eclipse/che-dotnet-6.0",
```
nickboldt, line 441: unless you're adding this image upstream too for Che 7.50 or 7.51, this test update is meaningless here.

nickboldt:
Note that in the current 3.x udi image:

```

$➔ podman run -it --rm --entrypoint /bin/bash quay.io/devspaces/udi-rhel8:3.2

bash-4.4# dotnet --version
6.0.106

bash-4.4# dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.106
 Commit:    e83a22f558

Runtime Environment:
 OS Name:     rhel
 OS Version:  8
 OS Platform: Linux
 RID:         rhel.8-x64
 Base Path:   /usr/lib64/dotnet/sdk/6.0.106/

Host (useful for support):
  Version: 6.0.6
  Commit:  7cca709db2

.NET SDKs installed:
  3.1.420 [/usr/lib64/dotnet/sdk]
  6.0.106 [/usr/lib64/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.26 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.6 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.26 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.6 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```

devspaces-udi/Dockerfile
omajid: Most users will target a specific version. Applications that target 3.1 will not run against 6.0, without the developers explicitly specifying that through a project configuration, runtime arguments or environment variables: https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward

nickboldt:
New plan: re-add 3.1 and add a deprecation warning that points to https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core and the EOL date of Dec 13 2022

omajid:
PR looks good to me!

svor:
Will 3.1 be a default version?

devspaces-udi/Dockerfile
```diff
@@ -26,7 +26,7 @@ ENV \
     E2FSPROGS_VERSION="1.46.5" \
     LD_LIBRARY_PATH="/usr/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}" \
     CPATH="/usr/include${CPATH:+:${CPATH}}" \
-    DOTNET_RPM_VERSION=3.1 \
+    DOTNET_RPM_VERSION="6.0" \
```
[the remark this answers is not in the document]
nickboldt: No idea how this env var is actually used, other than when I collect strings with "_VERSION" to pass the list to prod sec manifest.

```
grep -E -i "FROM|yum|rh-|INSTALL|COPY|ADD|curl|_VERSION"
```

https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/product/manifest/3.0.0/manifest_log.txt

We could change that to
```
DOTNET_RPM_VERSIONS="6.0 3.1" 
```
if we want to be marginally more acurate.

nickboldt:
> Will 3.1 be a default version?

no, it'll be 6.0. see https://github.com/redhat-developer/devspaces-images/pull/262#issuecomment-1178998309.

I've re-added the `dotnet` rpm, which I think sets up defaults/path/envs.

PR #264, opened 2022-07-04
chore: don't copy yarn.lock

svor:

Don't copy non existing files

Related issue: https://issues.redhat.com/browse/CRW-3110


nickboldt:
Caused by cleanup work in https://github.com/redhat-developer/devspaces/pull/752/files

PR #291, opened 2022-07-16
Set namespace: openshift-devspaces in CheCluster

dmytro-ndp:
There is wrong default namespace in CheCluster CRD: "devspaces"
Correct one: "openshift-devspaces".

### DIFF
--- a/devspaces-operator-bundle/manifests/devspaces.csv.yaml
+++ b/devspaces-operator-bundle/manifests/devspaces.csv.yaml
@@ -23,7 +23,7 @@ metadata:
           "kind": "CheCluster",
           "metadata": {
             "name": "devspaces",
-            "namespace": "devspaces"
+            "namespace": "openshift-devspaces"
           },
           "spec": {
             "auth": {


tolusha:
There several nuances: 
1. Updating `csv` have to be done via `sync-che-olm.sh` files [1] 
2. Namespace name in CheCluster CR sample is anyway ignored by OpenShift UI when user creates an instance.

[1] https://github.com/redhat-developer/devspaces-images/blob/365c25d1d84af92fd2878777f7eeb71261ad739f/devspaces-operator-bundle/build/scripts/sync-che-olm.sh

nickboldt:
Here's a fix for the CSV as applied by the sync*.sh script:

https://github.com/redhat-developer/devspaces-images/pull/293

Tested with:

```
cd /home/nboldt/2/devspaces-operator-bundle # (where I have git@github.com:redhat-developer/devspaces-images.git checked out)
./build/scripts/sync.sh -v 3.2.0 -s ~/0/op/ -t `pwd`
```

PR #295, opened 2022-07-19
Update from v2.5.0 to v2.8.1, see what happens. 

SDawley:
Done in an attempt further https://issues.redhat.com/browse/CRW-3136 (Migrate to Cachito).


ibuziuk:
I think we should update upstream first to avoid a mismatch between project and product

SDawley:
OK, what are you referring to when you say upstream? The only upstream I know of is https://github.com/traefik/traefik since that's what we pull into devspaces-traefik.

PR #311, opened 2022-08-09
CRW-3135 Update sync 

SDawley:
Update sync script to get dockerfiles from upstream (and include updated files from running locally).


devspaces-configbump/Dockerfile
```diff
@@ -48,7 +48,7 @@ LABEL summary="$SUMMARY" \
       io.openshift.tags="$PRODNAME,$COMPNAME" \
       com.redhat.component="$PRODNAME-$COMPNAME-container" \
       name="$PRODNAME/$COMPNAME" \
-      version="3.2" \
```
nickboldt, line 51: this should be version 3.2 in the 3.x branch.

devspaces-configbump/build/scripts/sync.sh
```diff
@@ -54,17 +54,15 @@ echo ".github/
...
 tests/
 /container.yaml
 /content_sets.*
 /cvp.yml
 /cvp-owners.yml
 get-source*.sh
 tests/basic-test.yaml
-sources
 make-release.sh
-rhel.Dockerfile
```
nickboldt, line 67: this means we will have a verbatim copy of rhel.Dockerfile in midstream, but the sync.sh script doesn't use it ... it uses the brew.Dockerfile (with some transformations).

So this is probably fine, as anyone who wants to build the midtream configbump for DS 3.x can do so from rhel.Dockerfile, which is consistent with the registries.

PR #314, opened 2022-08-17
fix: Remove unneeded GOPATH for UDI image

mkuznyetsov:

this GOPATH definition is not needed, since we rely on default GOPATH pointing to /home/user/go


svor (approved):
LGTM

PR #318, opened 2022-09-06
Initial checkin updating to use cachito in brew insead of sources build.

SDawley:
For: https://issues.redhat.com/browse/CRW-3136

- Updated container.yaml with cachito info (needed 2 sources, the main go dependencies and the webui yarn dependencies).
- removed sources file, rhel.binary.Dockerfile and rhel.binary.build.sh since they're no longer used.
- renamed rhel.container.Dockerfile to rhel.Dockerfile to be more consistent with other images.
- made get-sources.sh a force build script.


SDawley:
@nickboldt I'm sure I've forgotten something but I've forgotten what.

devspaces-traefik/build/rhel.Dockerfile
```diff
@@ -0,0 +1,46 @@
+# https://registry.access.redhat.com/ubi8-minimal 
+FROM registry.access.redhat.com/ubi8-minimal:8.6-902.1661794353 
+USER 0
+
+# cachito
```
nickboldt, line 5: none of these steps will work in midstream... which is OK, but not exactly "community-first" :(

devspaces-traefik/build/rhel.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+COPY $REMOTE_SOURCES $REMOTE_SOURCES_DIR
+RUN source $REMOTE_SOURCES_DIR/devspaces-images-traefik/cachito.env
+WORKDIR $REMOTE_SOURCES_DIR/devspaces-images-traefik/app/devspaces-traefik
+
+# cachito:yarn step 2: workaround for yarn not being installed in an executable path
+RUN ln -s $REMOTE_SOURCES_DIR/devspaces-images-traefik/app/devspaces-dashboard/.yarn/releases/yarn-*.js /usr/local/bin/yarn 
+
+RUN microdnf -y install dnf && \
+    dnf install -y golang python2 make gcc-c++ openssl-devel && \
+    dnf module install -y nodejs:12:8020020200326104117/development && \
```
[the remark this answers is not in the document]
SDawley: Yes. According to osbs docs that I cannot find now there is an issue with nodejs 12 specifically that can be worked around by using _that specific version_ and I need node 12 specifically and it was being very finicky

devspaces-traefik/build/rhel.Dockerfile
SDawley: Is it the source container or the user that's problematic? Or both?  ubi8-minimal:8.6-902.1661794353  is what we were using in rhel.container.Dockerfile, but I also used the more generic ubi8-minimal:8.6-902 which worked. 

rhel.binary.Dockerfile used registry.redhat.io/ubi8/nodejs-14:1-83 which was less happy. 

I remember there was a problem that was fixed by setting user 0 but I forget what so I could take it out and see if it breaks again

devspaces-traefik/build/rhel.Dockerfile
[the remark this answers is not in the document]
nickboldt: perhaps because node 12 is EOL / deprecated... and that one gets around the CVP failures? if so, winner.

devspaces-traefik/build/rhel.Dockerfile
nickboldt: my comments wasn't about the base image, it was about the copy commands that only work w/ cachito. but I'd rather make Legal happy than the community so... onward!

PR #333, opened 2022-09-23
Introduce local Openvsx in the plugin registry

svor:
### What does this PR do?
Add local openvsx

notes:

- it uses gradle wrapper to build the sources of openvsx server
- postgresql was installed from https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
- final user is **postgres** to make it possible to start postgresql

### What issues does this PR fix or reference?
https://issues.redhat.com/browse/CRW-3295


devspaces-pluginregistry/Dockerfile
```diff
@@ -11,96 +11,121 @@
...
-    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
-
-RUN sed -i /etc/httpd/conf.d/ssl.conf \
-    -e "s,SSLProtocol all -SSLv2,SSLProtocol all -SSLv3," \
-    -e "s,SSLCipherSuite HIGH:MEDIUM:!aNULL:!MD5,SSLCipherSuite HIGH:!aNULL:!MD5,"
-
-RUN sed -i /etc/httpd/conf/httpd.conf \
+RUN mkdir -p /mnt/rootfs
+# Install httpd and postgresql
+RUN yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
```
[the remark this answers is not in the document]
nickboldt: basically assume anything with a `curl` or an `https://` is not allowed :D

devspaces-pluginregistry/Dockerfile
```diff
@@ -11,96 +11,121 @@
...
     -e "s,logs/error_log,/dev/stderr," \
     -e "s,logs/access_log,/dev/stdout," \
     -e "s,AllowOverride None,AllowOverride All," && \
-    chmod a+rwX /etc/httpd/conf /run/httpd /etc/httpd/logs/
+    chmod a+rwX /mnt/rootfs/etc/httpd/conf /mnt/rootfs/etc/httpd/conf.d /mnt/rootfs/run/httpd /mnt/rootfs/etc/httpd/logs/
+
 STOPSIGNAL SIGWINCH
 
+RUN mkdir -m 777 /mnt/rootfs/var/www/html/v3
+COPY /build/scripts/import_vsix.sh /mnt/rootfs/usr/local/bin
```
nickboldt, line 76: you're copying from `/build/scripts/` here, and `./build/scripts/` on line 83. Intentional difference? Do we need BOTH copies to two different places?

devspaces-pluginregistry/Dockerfile
```diff
@@ -11,96 +11,121 @@
...
+COPY /build/scripts/start_services.sh /mnt/rootfs/usr/local/bin/
+RUN chmod 755 /mnt/rootfs/usr/local/bin/*.sh
+COPY /build/dockerfiles/openvsx.conf /mnt/rootfs/etc/httpd/conf.d/
+
+COPY README.md .htaccess /mnt/rootfs/var/www/html/
+
+COPY ./build/scripts/*.sh /build/
```
nickboldt, line 83: these 3 could be a single line as they all go to the same target folder.

devspaces-pluginregistry/Dockerfile
```diff
@@ -11,96 +11,121 @@
...
+    /usr/local/bin/import_vsix.sh && \
+    # add permissions for anyuserid
+    chgrp -R 0 /var/lib/pgsql/13/data/database && \
+    #cleanup postgresql pid
+    rm /var/lib/pgsql/13/data/database/postmaster.pid && \
+    rm /var/run/postgresql/.s.PGSQL* && \
+    rm /tmp/.s.PGSQL* && \
+    chmod -R g+rwX /var/lib/pgsql/13/data/database && mv /var/lib/pgsql/13/data/database /var/lib/pgsql/13/data/old
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
-CMD ["/usr/local/bin/rhel.entrypoint.sh"]
```
nickboldt, line 103: not needed anymore? should the file be deleted then, and not copied to the image on line 96?

devspaces-pluginregistry/.gitignore
```diff
@@ -23,6 +23,7 @@ v3/**/latest/
 resources/
 /root-local.tgz
 /resources.tgz
+openvsx-server.tar.gz
```
nickboldt, line 26: there are now 4 tarballs to ignore:
```
root-local.tgz resources.tgz openvsx-server.tar.gz nodejs.tar.gz
```

devspaces-pluginregistry/get-sources.sh
```diff
@@ -92,8 +92,25 @@ if [[ ${PULL_ASSETS} -eq 1 ]]; then
...
 	${BUILDER} rm -f pluginregistryBuilder
-
 	${BUILDER} rmi ${tmpContainer}
+
+	rm -f ./openvsx-server.tar.gz
+	OPENVSX_BUILDER_IMAGE=che-openvsx:latest
+	${BUILDER} build --progress=plain -f build/dockerfiles/openvsx-builder.Dockerfile \
+		-t "$OPENVSX_BUILDER_IMAGE" . --target builder
+	${BUILDER} create --name openvsxBuilder ${OPENVSX_BUILDER_IMAGE}
+	${BUILDER} cp openvsxBuilder:/openvsx-server.tar.gz .
```
nickboldt, line 102: I vaguely recall that `podman cp` doesn't work as expected in Jenkins so we might need a different approach here... but willing to try it and be proven wrong :D

devspaces-pluginregistry/build/scripts/sync.sh
```diff
@@ -106,11 +106,8 @@ sed "${TARGETDIR}/build/dockerfiles/Dockerfile" --regexp-extended \
     `# Strip registry from image references` \
```
nickboldt, line 106: add this to fix midstream refs to ubi8 so it works in Brew:

```suggestion
  `# Update ubi8 image name` \
  -e "s#ubi8/ubi:#ubi8:#g" \
  `# Remove registry so build works in Brew` \
  -e "s#FROM (registry.access.redhat.com|registry.redhat.io)/#FROM #g" \
```

devspaces-pluginregistry/Dockerfile
nickboldt, line 29: this won't work in Brew. See https://github.com/redhat-developer/devspaces/pull/797/files#r978573048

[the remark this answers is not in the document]
nickboldt: See https://github.com/redhat-developer/devspaces/pull/797/files#r978573048

devspaces-pluginregistry/Dockerfile
svor: yes, here we copy scripts that will be used at runtime. Another copy uses in the builder that won't affect the final image

nickboldt: right but one is `/build` and the other `./build` ... ?

svor: true, fixed

devspaces-pluginregistry/Dockerfile
svor: fixed

devspaces-pluginregistry/Dockerfile
```diff
@@ -11,96 +11,121 @@
...
+RUN tar -xvf /build/resources.tgz -C /build/
+RUN rm -rf /build/output/v3/che-theia-plugins.yaml /build/output/v3/che-editors.yaml /build/output/v3/che-plugins.yaml
+RUN /build/swap_plugins_memory.sh /build/output/v3
+RUN /build/list_referenced_images.sh /build/output/v3 --use-generated-content > /build/output/v3/external_images.txt && cat /build/output/v3/external_images.txt
+RUN chmod -R g+rwX /build
+
+RUN cp -r /build/output/v3 /mnt/rootfs/var/www/html/
+COPY v3/plugins/ /mnt/rootfs/var/www/html/v3/plugins/
+COPY v3/images/*.png /mnt/rootfs/var/www/html/v3/images/
+COPY ./build/dockerfiles/rhel.entrypoint.sh ./build/dockerfiles/entrypoint.sh /mnt/rootfs/usr/local/bin/
```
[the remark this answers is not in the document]
svor: rhel.entrypoint.sh was removed

devspaces-pluginregistry/.gitignore
svor: fixed in parent .gitignore

devspaces-pluginregistry/Dockerfile
svor: added a tarball for postgresql

devspaces-pluginregistry/get-sources.sh
```diff
@@ -59,7 +59,7 @@ if [[ ${PULL_ASSETS} -eq 1 ]]; then
 		--target builder --build-arg BOOTSTRAP=true
 	echo "<======= END BOOTSTRAP BUILD ======="
 	# update tarballs - step 2 - check old sources' tarballs
-	TARGZs="root-local.tgz resources.tgz"
+	TARGZs="root-local.tgz resources.tgz openvsx-server.tar.gz nodejs.tar.gz"
```
nickboldt: do we need the postgres tarball here too?

devspaces-pluginregistry/get-sources.sh
svor: we've already had a similar piece of code: https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-pluginregistry/get-sources.sh#L92-L96
was it a problem?

devspaces-pluginregistry/build/scripts/sync.sh
```diff
@@ -106,11 +106,8 @@ sed "${TARGETDIR}/build/dockerfiles/Dockerfile" --regexp-extended \
...
     -e 's|FROM registry.redhat.io/|FROM |' \
-    `# CRW-2448 switch from ubi8 to rhel8 for OSBS` \
-    -e 's|ubi8/httpd-24:([0-9]+)(-[0-9.]+)|rhel8/httpd-24:\1|g' \
-    -e 's|ubi8/httpd-24$|rhel8/httpd-24|g' \
     `# trim off version so we get the latest from internal registry` \
-    -e 's|ubi8/python-38:([0-9]+)(-[0-9.]+)|ubi8/python-38:\1|g' \
+    -e 's|ubi8/ubi:([0-9]+)(-[0-9.]+)|ubi8/ubi:\1|g' \
     `# Set arg options: disable BOOTSTRAP; update DS_BRANCH to correct value` \
     -e 's|ARG BOOTSTRAP=.*|ARG BOOTSTRAP=false|' \
     -e "s|ARG DS_BRANCH=.*|ARG DS_BRANCH=${DS_BRANCH}|" \
```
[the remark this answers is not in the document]
svor: some code was reverted to install **yq** and the script was updated

devspaces-pluginregistry/get-sources.sh
nickboldt: nope, if we're already doing that, we're good. Just checking / ignore me / "I'm new, it's my first day."

PR #344, opened 2022-11-02
CRW-3138: Update Opertator to use Cachito

SDawley:
Add new dockerfile for brew and update sync to use that.
Update container.yaml with additional sources.


nickboldt (changes requested):
Does a change need to happen in ./build/scripts/util.sh too? it's called by ./get-sources.sh and ./build/scripts/sync.sh and defines a method `updateDockerfileArgs` that updates the Dockerfile...

SDawley:
> Does a change need to happen in ./build/scripts/util.sh too? it's called by ./get-sources.sh and ./build/scripts/sync.sh and defines a method `updateDockerfileArgs` that updates the Dockerfile...

get-sources no longer references util.sh and updateDockerfileArgs isn't called by sync.sh anymore as the versions are defined in job-config.json so I can probably remove the script entirely.

nickboldt:
>  probably remove the script entirely

as long as nothing else uses it, then yeah. Simplication!

nickboldt (approved):

PR #388, opened 2023-01-09
fix: alm-example, headerMessage.show must be bool

tolusha:

### Reference issue 
https://issues.redhat.com/browse/CRW-3785


dmytro-ndp (approved):
LGTM

PR #392, opened 2023-01-16
chore: don't add lombok.jar into UDI

svor:

Since VSCode redhat.java extension doesn't require the path to lombok.jar we can avoid adding lombok.jar into UDI 

Related issue: https://issues.redhat.com/browse/CRW-3602


nickboldt:
Seems reasonable but I would not merge this until we know FOR SURE that we'll be dropping theia from the 3.5 payload. 

@Kasturi1820 FYI, this is one of the several changes that will be queued to include in 3.5 which we can't merge until we :100: decide when Theia goes away.

nickboldt:
> should I do additional cleanup somewhere?

I think you've done all that's needed: 
* don't fetch the jar in yaml
* don't use the jar in dockerfile

nickboldt (approved):

PR #402, opened 2023-02-16
Moving dockerfiles to match upstream and editing sync.sh

SDawley:
Sync.sh should handle new location and brew.Dockerfile now.

Depends on https://github.com/che-incubator/kubernetes-image-puller/pull/125


devspaces-imagepuller/build/scripts/sync.sh
```diff
@@ -72,7 +72,7 @@ rm -f /tmp/rsync-excludes
 find ${TARGETDIR}/ -name "*.sh" -exec chmod +x {} \;
 
 # transform Dockerfile
```
nickboldt, line 74: can this whole block be removed, since the brew.Dockerfile is already correct upstream, minus the addtion of the envs and label on line 86?

nickboldt:
related to JIRA:
https://issues.redhat.com/browse/CRW-4043

PR #406, opened 2023-02-23
CRW-4043 Add brew.Dockerfile to devspaces-idea

SDawley:


nickboldt (approved):

devspaces-idea/projector.sh
```diff
@@ -305,7 +305,7 @@ EOM
     docker build \
     --progress="$PROGRESS" \
     -t "$CONTAINER_TAG" \
-    -f Dockerfile .
+    -f ./build/dockerfiles/brew.Dockerfile .
```
RomanNikitenko, line 308: I'm afraid this change will be overwritten by upstream change at sync step...
So, after synchronization it will be https://github.com/che-incubator/jetbrains-editor-images/pull/160/files#diff-2f8c6b58a5febd98b78c35887f0c751d6aa853cb2dbabb4e3be896f695a9d41dR308

Samantha, could you check it?

devspaces-idea/che-plugin/.dockerignore
```diff
@@ -1 +1 @@
-Dockerfile
+build/dockerfiles/brew.Dockerfile
```
RomanNikitenko, line 1: the same as for https://github.com/redhat-developer/devspaces-images/pull/406/files#r1126195967
I guess it will be overwritten by upstream change: 
`build/dockerfiles/brew.Dockerfile` ==> `build/dockerfiles/Dockerfile`
https://github.com/che-incubator/jetbrains-editor-images/pull/160/files#diff-9320adfbe6de51ccbec6764e05304d95b3cb6f58d52556dc9003f951fd7e7632R1

devspaces-idea/build/dockerfiles/brew.Dockerfile
```diff
@@ -92,20 +92,4 @@ USER 1001
...
-      io.k8s.display-name="$DESCRIPTION" \
-      io.openshift.tags="$PRODNAME,$COMPNAME" \
-      com.redhat.component="$PRODNAME-$COMPNAME-container" \
-      name="$PRODNAME/$COMPNAME" \
-      version="3.6" \
-      license="EPLv2" \
-      maintainer="Vladyslav Zhukovskyi <vzhukovs@redhat.com>" \
-      io.openshift.expose-services="" \
-      usage=""
+# append Brew metadata here
```
RomanNikitenko, line 95: `brew.Dockerfile` should come to the current repo at the synchronization step

devspaces-idea/projector.sh
SDawley: That's a very good question because if we move Dockerfile to ./build/dockerfiles/Dockerfile projector.sh will be correct in che/images but we'll need to change it to ./Dockerfile for the brew build...

devspaces-idea/build/dockerfiles/brew.Dockerfile
nickboldt: +1, if you're cool with having "downsteam useful" artifacts like rhel.Dockerfile and brew.Dockerfile in the upstream repo.

PR #416, opened 2023-03-16
Chore: CRW-4043 

SDawley:
Move dockerfiles into build/dockerfiles and update sync.sh


nickboldt (approved):
Seems ok, but please make sure that a complete 3.8 traefik build is successful before resolving this part of CRW-4043.

PR #452, opened 2023-05-11
chore: don't exclude IDEA in the dashboard editor's view

svor:
Now we want to show che-idea editor in the User Dashboard, issue about that:

https://issues.redhat.com/browse/CRW-4139

Also we don't need to exclude che-theia editor since it was removed from the plugin registry and it doesn't exist at all


nickboldt (changes requested):
This change directly contradicts https://issues.redhat.com/browse/CRW-3485

nickboldt (approved):
+1, but I'd like to see a statement from CEE/PM in the JIRA stating we're undoing the decision in 3.3 intentionally.

PR #459, opened 2023-06-08
[WIP] feat: add devfile and vscode settings to the repo

mkuznyetsov:

Add basic devfile structure and vscode extension settings
WIP points
1. add commands to build every projects container
2. add components (unless default component is good enough)


devfile.yaml
```diff
@@ -0,0 +1,41 @@
...
+#   Red Hat, Inc. - initial API and implementation
+#
+schemaVersion: 2.2.0
+metadata:
+  name: devspaces-images
+# add components:, unless basic udi component is good enough?
+components:
+  - name: tools
+    container:
+      image: quay.io/devspaces/devspaces-images-dev:latest
```
l0rd, line 19: When trying your PR on workspaces.openshift.com I get "Failed to pull image "quay.io/devspaces/devspaces-images-dev:latest": rpc error: code = Unknown desc = reading manifest latest in quay.io/devspaces/devspaces-images-dev: unauthorized: access to the requested resource is not authorized". Maybe the image is private or not pushed yet?

devfile.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+schemaVersion: 2.2.0
+metadata:
+  name: devspaces-images
+# add components:, unless basic udi component is good enough?
```
l0rd, line 15:
```suggestion
```
I think you can remove this no? If you have added the Dockerfile it's because you want go v1.18 (and currently basic udi has 1.19) right?

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+    "recommendations": [
+        "rogalmic.bash-debug",
+        "redhat.vscode-yaml"
```
l0rd, line 4: What about `golang/vscode-go`

devfile.yaml
mkuznyetsov: it needs to be 1.19, similarly to how it is defined in upstream. it's probably better to hardcode it this way, as udi's golang version can go further (it is at 1.20 right now)

.devfile.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+# This container featres tools required for Devspaces Images
+FROM quay.io/devfile/universal-developer-image:ubi8-latest
+
+#install Go 1.18 (needed to work with devspaces-operator)
+RUN cd /tmp && wget https://go.dev/dl/go1.18.6.linux-amd64.tar.gz && \
+        mkdir $HOME/go1.18 && \
```
nickboldt, line 16: can we update this to go 1.20 since that's what we're using downstream?

devfile.yaml
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) 2019-2021 Red Hat, Inc.
```
nickboldt, line 2: new file! old copywrong date.

devfile.yaml
```diff
@@ -0,0 +1,43 @@
...
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+schemaVersion: 2.2.0
+metadata:
+  name: devspaces-images
+components:
+  - name: tools
+    container:
+      image: quay.io/devfile/universal-developer-image:ubi8-latest
```
nickboldt, line 18: why are you using the upstream UDI image in a devfile for the downstream UDI image?

.devfile.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+# This container featres tools required for Devspaces Images
+FROM quay.io/devfile/universal-developer-image:ubi8-latest
```
nickboldt, line 12: if this is something you want to see built downstream, you aren't allowed to use `FROM quay.io` in OSBS.

mkuznyetsov: this image is not needed to be built downstream. It's only used as a tooling container, when this workspace will be used in Che/Devspaces

PR #465, opened 2023-06-22
feat: add quay.io/devspaces/ansible-creator-ee into related images list

svor:
add quay.io/devspaces/ansible-creator-ee image with digest into related images list


nickboldt:
Seems reasonable but I bet your operator-bundle build will fail in CVP testing because you refer to a quay.io image instead of a registry.redhat.io image. Feel free to merge and see what explodes. :)

nickboldt (approved):

PR #471, opened 2023-07-06
Provide Node.js for running VS Code

RomanNikitenko:
Provide Node.js for running VS Code.
We need this change for https://github.com/eclipse/che/issues/21778

### DIFF
--- a/devspaces-code/build/dockerfiles/libc-content-provider.Dockerfile
+++ b/devspaces-code/build/dockerfiles/libc-content-provider.Dockerfile
@@ -45,16 +45,46 @@ RUN yarn install --force
 
 RUN echo "$(ulimit -a)"
 
-# Compile
-RUN NODE_ARCH=$(echo "console.log(process.arch)" | node) \
+ARG PLATFORM="linux"
+ARG NODE_LOCATION="https://nodejs.org"
+
+RUN { \ 
+    if [[ $(uname -m) == "s390x" ]]; then \
+      NODE_ARCH="s390x"; \
+      BUILD_ARCH="s390x"; \
+    elif [[ $(uname -m) == "ppc64le" ]]; then \ 
+      NODE_ARCH="ppc64le"; \
+      BUILD_ARCH="ppc64"; \
+    elif [[ $(uname -m) == "x86_64" ]]; then \
+      NODE_ARCH="x64"; \
+      BUILD_ARCH="x64"; \
+    elif [[ $(uname -m) == "aarch64" ]]; then \
+      NODE_ARCH="arm64"; \
+      BUILD_ARCH="arm64"; \
+    else \
+      NODE_ARCH=""; \
+      BUILD_ARCH=$(echo "console.log(process.arch)" | node); \
+    fi; \ 
+    } \
     && NODE_VERSION=$(cat /checode-compilation/remote/.yarnrc | grep target | cut -d ' ' -f 2 | tr -d '"') \
-    # cache node from this image to avoid to grab it from within the build
-    && mkdir -p /checode-compilation/.build/node/v${NODE_VERSION}/linux-${NODE_ARCH} \
-    && echo "caching /checode-compilation/.build/node/v${NODE_VERSION}/linux-${NODE_ARCH}/node" \
-    && cp /usr/bin/node /checode-compilation/.build/node/v${NODE_VERSION}/linux-${NODE_ARCH}/node \
-    && NODE_OPTIONS="--max_old_space_size=8500" ./node_modules/.bin/gulp --tasks \
-    && NODE_OPTIONS="--max_old_space_size=8500" ./node_modules/.bin/gulp vscode-reh-web-linux-${NODE_ARCH}-min \
-    && cp -r ../vscode-reh-web-linux-${NODE_ARCH} /checode
+    && { \
+        if [ -n "$NODE_ARCH" ]; then \
+            NODE_URL="${NODE_LOCATION}/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-${PLATFORM}-${NODE_ARCH}.tar.gz"; \
+            echo "Downloading Node.js from ${NODE_URL}"; \
+            wget -q "${NODE_URL}"; \
+            tar -xf node-v${NODE_VERSION}-${PLATFORM}-${NODE_ARCH}.tar.gz; \
+            mv node-v${NODE_VERSION}-${PLATFORM}-${NODE_ARCH} nodejs; \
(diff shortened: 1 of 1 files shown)


devspaces-code/build/dockerfiles/libc-content-provider.Dockerfile
```diff
@@ -45,16 +45,46 @@ RUN yarn install --force
...
-    && echo "caching /checode-compilation/.build/node/v${NODE_VERSION}/linux-${NODE_ARCH}/node" \
-    && cp /usr/bin/node /checode-compilation/.build/node/v${NODE_VERSION}/linux-${NODE_ARCH}/node \
-    && NODE_OPTIONS="--max_old_space_size=8500" ./node_modules/.bin/gulp --tasks \
-    && NODE_OPTIONS="--max_old_space_size=8500" ./node_modules/.bin/gulp vscode-reh-web-linux-${NODE_ARCH}-min \
-    && cp -r ../vscode-reh-web-linux-${NODE_ARCH} /checode
+    && { \
+        if [ -n "$NODE_ARCH" ]; then \
+            NODE_URL="${NODE_LOCATION}/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-${PLATFORM}-${NODE_ARCH}.tar.gz"; \
+            echo "Downloading Node.js from ${NODE_URL}"; \
+            wget -q "${NODE_URL}"; \
```
RomanNikitenko, line 74: @nickboldt 
I would like to know your opinion about current PR.
We need it for https://github.com/eclipse/che/issues/21778.

The main difference is: Node.js is going to be downloaded at the building `libc-content-provider.Dockerfile` step.

nickboldt: > Node.js is going to be downloaded at the building

Sounds like the opposite of what was done for https://issues.redhat.com/browse/CRW-3160 ... you can't download random stuff from the internet anymore, you can only use rpms :( 

Do you require a specific version of nodejs that's not available from UBI 8 or UBI 9 rpm content sets?

RomanNikitenko: thank you for the quick answer!

> Do you require a specific version of nodejs that's not available from UBI 8 or UBI 9 rpm content sets?

`ubi-8` brings nodejs that incompatible with `ubi-9`.
I tested in the `che-code` that starting VS Code works well for both (ubi-8 and ubi-9) when Node.js is downloaded...

RomanNikitenko: One more question:
in the upstream we can use [From docker.io/node:16.17.1-alpine3.15](https://github.com/che-incubator/che-code/blob/d001c152fc445c195b46433fb12b363e5be63488/build/dockerfiles/linux-musl.Dockerfile#L10)

If I'm not mistaken we can not use  `From docker.io/node:16.17.1-alpine3.15` for the `Devspaces-Code`.
Do you know a workaround for such use case?

RomanNikitenko: > Side question: why are we using Node 16 when 18 is available?

It's VS Code requirement: https://github.com/microsoft/vscode/wiki/How-to-Contribute#prerequisites

nickboldt: ok, well https://catalog.redhat.com/software/containers/ubi8/nodejs-16/615aee9fc739c0a4123a87e1?container-tabs=packages contains `nodejs-16.19.1-1.module+el8.7.0+18373+704f5cef.x86_64` ... which should be compatible with required [Node.JS](https://nodejs.org/en/), x64, version `>=16.17.x and <17`

RomanNikitenko: > Are you saying these versions don't work for your build?

Let's say I've built `libc` docker image from the dockerfile based on the `ubi-8` - it contains Node.js 16 that requires `OpenSSL 1`.
- we run VS Code in the `ubi-8` based container - it works - no problem!
- we run VS Code in the `ubi-9` based container - it doesn't work - `ubi-9` image uses `OpenSSL 3`

The same problem when I build `libc` docker image from the dockerfile based on the `ubi-9`, but user tries to run VS Code in the `ubi-8` based container!

I don't have such problems when I just download required version of Node.js - the workspace can be started in both: `ubi-8` and `ubi-9` based containers...

nickboldt:
I wonder if the libc build can be set up to NOT directly depend on a specific openssl implementation, or to include its own embedded one? 

> I don't have such problems when I just download required version of Node.js

Are you downloading the 66MB [node-v16.20.1.tar.gz](https://nodejs.org/download/release/latest-v16.x/node-v16.20.1.tar.gz) or the 33MB [node-v16.20.1-linux-x64.tar.gz](https://nodejs.org/download/release/latest-v16.x/node-v16.20.1-linux-x64.tar.gz) ? 

Looks like the larger tarball includes: 
* node-v16.20.1/deps/openssl/
* node-v16.20.1/deps/openssl/openssl/util/opensslwrap.sh
* arch-specific config, header, and node-gyp files 

But the smaller one only includes:
* node-v16.20.1-linux-x64/include/node/openssl/ *.h headers (for multiple arches, which is odd for an arch-specific tarball (?))

So... maybe it's the openssl wrapper that's solving your problem for RHEL 8 vs. 9? Both tarballs include these:

* node-v16.20.1/deps/openssl/config/archs/linux-x86_64/no-asm/include/openssl/opensslconf.h
* node-v16.20.1/deps/openssl/config/archs/linux-x86_64/asm_avx2/include/openssl/opensslconf.h
* node-v16.20.1/deps/openssl/config/archs/linux-x86_64/asm/include/openssl/opensslconf.h

* node-v16.20.1-linux-x64/include/node/openssl/archs/linux-x86_64/no-asm/include/openssl/opensslconf.h
* node-v16.20.1-linux-x64/include/node/openssl/archs/linux-x86_64/asm_avx2/include/openssl/opensslconf.h
* node-v16.20.1-linux-x64/include/node/openssl/archs/linux-x86_64/asm/include/openssl/opensslconf.h

Alternatively, perhaps it's time to migrate Dev Spaces to RHEL 9 so we don't have to worry about RHEL 8 anymore. (This will take a few releases to achieve, however, but is ultimately where DS will end up.)

nickboldt:
Note that a dependency on libssl or libcrypto indicates that there are FIPS implications here. We need to be careful that the node we're using is FIPS-friendly. Static links are generally not cool. See https://issues.redhat.com/browse/CRW-4582?focusedId=22617302&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22617302

PR #496, opened 2023-08-30
feat: CRW-3160 devspaces-code build with Cachito

mkuznyetsov:
Changes for midstream to facilitate devspaces-code with Cachito build https://issues.redhat.com/browse/CRW-3160 :

- use project sources and yarn dependencies provided by Cachito (`container.yaml`, `fetch-artifacts-url.yaml`), instead of assets tar.gz files
- added brew.Dockerfile with Cachito specific configuration https://issues.redhat.com/browse/CRW-4043 
   - Because of rather different approach for Cachito build in downstream, where dockerfile contains significant differences, like compile stage is a based on former [build image](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-code/build/dockerfiles/libc-content-provider.Dockerfile), as well as stage for building machinexec, difference in testing steps, and final assembly step. 
   I believe it will be simpler for developers to use and maintain two Dockerfiles (for `local` and `brew` builds), instead of attempting to combine and transform `local` dockerfile into `brew` with numerous scripts, that developers would have to be aware of.
- add Dockerfile for local build of midstream https://issues.redhat.com/browse/CRW-4043
- sync.sh script will process brew.Dockerfile, updating list of modules in `container.yaml` and brew.Dockerfile, update checksums in fetch-artifacts-url.yaml for vsix files.


nickboldt:
Converted to draft as the description has WIP. LMK when this is ready for review. Also needs to be updated to the latest changes in the branch.

devspaces-code/build/dockerfiles/brew.Dockerfile
```diff
@@ -0,0 +1,277 @@
...
+
+# hadolint ignore=SC3045
+RUN echo "$(ulimit -a)"
+
+
+# Cachito clears all project's '.yarnrc' files. To make sure yarn is configured to the local Nexus.
+# But VS Code build checks some fields in the '.yarnrc' and fails if it's empty.
+# So, we need to restore the fields before running the build. The actual values are ignored in our case.
+# https://github.com/microsoft/vscode/blob/bceaaf84a27c3a95a0cdfc79287e3215b56b951c/build/gulpfile.reh.js#L128
+RUN echo 'target "19.1.8"' > .yarnrc; \
```
azatsarynnyy, line 173: I think we can try to skip this (to simplify a bit) as we're creating the folder that VS Code expects and copying the Node version that is in the image
https://github.com/redhat-developer/devspaces-images/pull/496/files#diff-00b3f0589c23db54e78e6753796ea673aba5e2947f31914dde7c8c920ec5cfd9R181-R187

devspaces-code/build/scripts/sync.sh
```diff
@@ -12,6 +12,38 @@
...
 set -e
 
+BASE_URL="https://download.devel.redhat.com/rcm-guest/staging/devspaces/build-requirements"
+
+generateFetchArtifactsURLYaml () {
+  # Download manifest if it's not there
+  if [[ ! -f "/tmp/plugin-manifest.json" ]]; then
+    curl -sSL https://raw.githubusercontent.com/redhat-developer/devspaces-vscode-extensions/devspaces-3-rhel-8/plugin-manifest.json --output /tmp/plugin-manifest.json
+  else
+    echo "manifest detected"
```
nickboldt, line 22: is this a warning? an error? why not provide the path to the file so it's clear there was leftover trash from a previous run... or for that matter, should you delete `/tmp/plugin-manifest.json` before fetching it so it's always fresh? 

Also should this curl be aware of the branch being built so that we can have a different plugin-manifest.json for 3.8, 3.9, 3.x... ?

devspaces-code/build/scripts/sync.sh
```diff
@@ -66,6 +99,7 @@ code/src/vs/platform/keyboardLayout/electron-main/keyboardLayoutMainService.ts
 /cvp.yml
 /cvp-owners.yml
 devfile.yaml
+fetch-artifacts-url.yaml
```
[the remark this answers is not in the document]
nickboldt: I think the point is that upstream should not accidentally overwrite fetch-artifacts-url.yaml in midstream during a `rsync`. 

But yes, there's no point having this file upstream as Che can't use it... only DS can. 

> Would it be better to have a file for the ripgrep SHAs that could be parsed when generating the fetch-artifacts file so there was less manual work?

You mean instead of manually fetching the ripgrep binaries we'd automat this with:
* query GH API for new releases of https://github.com/microsoft/ripgrep-prebuilt/releases
* download the binaries we need, storing SHAs in some file, and upload that onto download.devel
* generate fetch-artifacts-url.yaml every time using the tarballs/SHAs on download.devel
* manually delete old versions of ripgrep prebuilt when they're no longer needed by 3.prev, 3.latest, or 3.next builds ?

devspaces-code/build/scripts/sync.sh
```diff
@@ -91,6 +125,53 @@ sed_in_place() {
...
+  # in case it's not gonna be in the end of the file anymore
+  # e.g. with yq added in for loop
+  # yq -y --in-place  '.remote_sources[0].remote_source.packages.yarn += { "path": "'$yarn_module'" }'
+  # but yq removes comments and formatting
+  sed_in_place '0,/this section is automatically generated in sync.sh/!d' ${TARGETDIR}/container.yaml
+  echo -e "$CONTAINER_YAML_MODULE_LIST" >> ${TARGETDIR}/container.yaml
+  # --- END update container.yaml and brew.Dockerfile
+
+  # --- BEGIN fetch-artifacts-url.yaml - generate list checksums
+  # fetch artifacts yaml consists of two parts - ripgrep libraries (that are uploaded to rcm-tools manually),
```
nickboldt, line 171: To Sam's point elsewhere, this could be automated (see other comment) but considering how infrequently there's a new https://github.com/microsoft/ripgrep-prebuilt/releases available, the effort to automate is probably not worth it. I like this half-manual, half-automated approach to keeping this file up to date.

devspaces-code/build/scripts/sync.sh
mkuznyetsov: good point, it should be downloaded every time, and I'll make the branch parametrized too

mkuznyetsov: now downloading files from branch, that the repo   (where sync.sh is checked out from) is checked out to, just like similar files job-config.json are fetched in sync.sh

nickboldt: Not seeing the branch param... looks like `devspaces-3-rhel-8` is still hardcoded in the above snippet.

mkuznyetsov: sorry, left a comment before pushing a rebased branch featuring the change

nickboldt:
still a couple of unresolved comments but you're getting closer!

devspaces-code/build/scripts/sync.sh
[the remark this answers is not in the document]
mkuznyetsov: for the first point - yes. the fetch-artifacts-url.yaml file is here added to the exclude list, so that rsync --delete will not delete this file in target (midstream), since the file is not present in the upstream. It's similarly to how it is done with other fetch-***.yaml files in our other midstream projects

@nickboldt for the latter point, I'm trying to extend the sync.sh to handle the update of builtins, in necessary
so for ripgrep, 
- I will check which version of vscode-ripgrep is used, and which version of ripgrep-prebuilt is needed for it [here](https://github.com/microsoft/vscode-ripgrep/blob/main/lib/postinstall.js#L21)
- then, if we have this prebuillt version already on rcm-tools - good, use its sha in .yaml, if not - then upload it using [devspaces script](https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/product/uploadBuildRequirementsToSpmmUtil.sh) and use its sha in .yaml
A manifest to track SHAs could look like this:
```
{
  "13.0.0-4": {
    "sha":"abc123",
    "source":"abc123"
  },
  "13.0.0-7": {
    "sha":"abc456",
    "source":"abc456"
  },
  "13.0.0-10": {
    "sha":"abc456",
    "source":"abc456"
  }
}
```
I suppose it will be sufficient enough to just fetch, update and rsync this manifest file directly to rcm utils, no need to have intermediary file in a git repository, like [here](  https://github.com/redhat-developer/devspaces-vscode-extensions/blob/devspaces-3-rhel-8/plugin-manifest.json ) ?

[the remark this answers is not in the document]
nickboldt: I only see one file for rigrep versions... but where are you storing which versions of ripgrep are needed for each arch and version of Dev Spaces?

Going forward we will need at least TWO ripgrep versions per devspaces release, because Z&P require an older one.

https://issues.redhat.com/browse/CRW-4938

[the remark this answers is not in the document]
nickboldt: Also just curious but why does it take 6 weeks for you to reply to PR comments?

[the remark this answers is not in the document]
mkuznyetsov:
```
{
  "13.0.0-7": {
    "powerpc64le": {
      "sha":"3413fb1941149842e712ba405c28e1c1d63d8cebe20d4cf7e6dc23ed9c1822ed",
      "source_sha":"e13d511ff6713cfb50a3e4f2eea793cb1a5489d6ae55d96e16a4698fd23a77f5"
    },
    "x86_64": {
      "sha":"23b9b87df7500572dc0e60d8db68dc7b91c7f4e3d21927ffade126f43dfd416b",
      "source_sha":"e13d511ff6713cfb50a3e4f2eea793cb1a5489d6ae55d96e16a4698fd23a77f5"
    },
    "s390x": {
      "sha":"facd18e0aff41d2fdb8364ff38d951aeb8ec0380e3010623f686425544a4ddde",
      "source_sha":"e13d511ff6713cfb50a3e4f2eea793cb1a5489d6ae55d96e16a4698fd23a77f5"
    }
  },
  "13.0.0-10": {
    "powerpc64le": {
      "sha":"40979af23debe10e33432ff98a49e30217c0a53c55affc3b3331ab2b2bc99f3f",
      "source_sha":"9c5d369ab6ad9b1af75bc89489d3bd47a9d2bff69449cbc4d6f288d7d96d9940"
    },
    "x86_64": {
      "sha":"ef820a62c1d6fdc396646762ff0f0e47e127947073ed8b5aa4ceea8b61cb1659",
      "source_sha":"9c5d369ab6ad9b1af75bc89489d3bd47a9d2bff69449cbc4d6f288d7d96d9940"
    },
    "s390x": {
      "sha":"61afc474f39ebe51424438c981a201ab1eed5ae55f3e82ff7561b6a8c9bd8b87",
      "source_sha":"9c5d369ab6ad9b1af75bc89489d3bd47a9d2bff69449cbc4d6f288d7d96d9940"
    }
  }
}
```
What if a scheme like this would be present on rcm-tools (but could be duplicated somewhere in git for clarity).
It doesn't say which DS version it's tied to, as it will be defined in DS version branch' sync.sh itself, which builtins we'd want to fetch.
So in the updated version of PR, it will fetch by default binaries for all platforms, and (if needed for CRW-4938, fetch extra versions per platform too, which then would have to be swapped during the build, if I understand it correctly)

devspaces-code/build/scripts/sync.sh
mkuznyetsov: like ripgrep-prebuilt, this will be handled in sync-builtins.sh

nickboldt:
Just a general comment... Commit messages that are just "fixup fixup fixup" aren't very meaningful. Please try to describe the changes being done to make it easier to understand the reason for the differences between versions of PRs.

PR #500, opened 2023-09-12
fix: Update devspace operator deployment labels

tolusha:
### Reference issue 
https://issues.redhat.com/browse/CRW-4848


nickboldt (approved):
does this affect the operator sync scripts? or just the csv? if the csv only, then approved

PR #501, opened 2023-09-22
Update podman-wrapper.sh access mode

l0rd:
That's a fix for https://issues.redhat.com/browse/CRW-4863


nickboldt:
issue description says "access mode" but all you're changing is the owner to root. Did you mean to `chmod +x` so this is an executable script?

nickboldt (approved):
ignore my comments, took a while to understand what this hackaround was doing. :D approved.

PR #503, opened 2023-09-22
Fix linking of ~/.java/current to proper java binaries

vitaliy-guliy:
On workspace startup it appeared `~/.java` directory is removed and then created again ( is is not exact that provided by universal developer image ), `~/.java/current` appeared to be missing and cannot be used to link the java binaries.

I've added as well a couple of `.gitignore` files to have more comfortable coding in VS Code startalone.

Solves https://issues.redhat.com/browse/CRW-4525

How to test
- create a workspace with https://github.com/vitaliy-guliy/demo-backend/tree/update-udi-to-checode
- open a terminal, check for java existence
- open `entrypoint.sh` from the root, check it contains similar with this PR files


devspaces-idea/projector-server/projector-server/.gitignore
```diff
@@ -0,0 +1 @@
+bin
```
nickboldt, line 1: You could have a `.gitignore` in the root of the `devspaces-idea/.gitignore` that ignores bin folders?, instead of having dozens of tiny files. Is there some case where you DON'T want git to ignore the bin folder?

devspaces-udi/etc/entrypoint.sh
```diff
@@ -54,15 +54,16 @@ fi
...
 if [ "${USE_JAVA8}" == "true" ] && [ ! -z "${JAVA_HOME_8}" ]; then
-  rm -r "${HOME}"/.java/current/*
   ln -s "${JAVA_HOME_8}"/* "${HOME}"/.java/current
   echo "Java environment set to ${JAVA_HOME_8}"
 elif [ "${USE_JAVA17}" == "true" ] && [ ! -z "${JAVA_HOME_17}" ]; then
-  rm -r "${HOME}"/.java/current/*
   ln -s "${JAVA_HOME_17}"/* "${HOME}"/.java/current
   echo "Java environment set to ${JAVA_HOME_17}"
 else
+  ln -s "${JAVA_HOME_11}"/* "${HOME}"/.java/current
```
nickboldt, line 66: could we add a TODO marker so we remember to switch to JDK 17 as default some time before OpenJDK 11 hits EOL in Oct 2024? https://access.redhat.com/articles/1299013

devspaces-idea/projector-server/projector-server/.gitignore
vitaliy-guliy: VS Code created those directories after I confirmed I trust this repository. Then I was proposed to commit around 180 files from that dirs.
So if binary directories were created on my laptop in the background, I would not commit the to the repository.

nickboldt: sure but https://github.com/redhat-developer/devspaces-images/commit/a0b48f91713de1b09bd1a9e912a89ab5a5124a4c achieves the same thing with one file instead of 16.

vitaliy-guliy: I think we have to add it one level upper, cause your commit was overwritten when syncing idea https://github.com/redhat-developer/devspaces-images/commits/devspaces-3-rhel-8/devspaces-idea/.gitignore

PR #521, opened 2023-11-06
Update rpms in machineexec and operator to fix C grades.

SDawley:


nickboldt:
your fixes for midstream 3.9 operator and machineexec have been overwritten by sync.sh reading upstream ... https://github.com/redhat-developer/devspaces-images/commits/devspaces-3.9-rhel-8

going to need to change that in che-machine-exec and che-operator repos, I suspect

https://github.com/eclipse-che/che-operator/blob/main/Dockerfile#L53
https://github.com/eclipse-che/che-machine-exec/blob/main/build/dockerfiles/brew.Dockerfile#L32 and https://github.com/eclipse-che/che-machine-exec/blob/main/build/dockerfiles/rhel.Dockerfile#L32

PR #525, opened 2023-11-14
Move tools & configs from /home/user/ to /home/tooling/

AObuchow:
This is a downstream port of https://github.com/devfile/developer-images/pull/115, that fixes these downstream issues [1](https://issues.redhat.com/browse/CRW-4972) and [2](https://issues.redhat.com/browse/CRW-5332), with some adaptations made to account for some differences between the upstream and downstream UDI image, namely:
- The lack of a Red Hat packaged version of GNU stow: we are instead using cachito to build stow from it's [GitHub mirror repo ](https://github.com/aspiers/stow)
- The use of bash's `-e` in the downstream UDI's entrypoint, which cause the entrypoint to fail if any command it executes has a non-zero return code
- The lack of sdkman and nvm on the downstream UDI
- The lack of vim on the downstream UDI

I also ported https://github.com/devfile/developer-images/pull/117 while I was at it in https://github.com/redhat-developer/devspaces-images/commit/afb8dfe9e579be811209a4f697e06c05d8a95744.

I tried to organize the commits in this PR to mirror their upstream equivalents. Below is a list of downstream to upstream commit conversions:
- https://github.com/redhat-developer/devspaces-images/commit/42fae05bf56b576cd5ec6924874707bb1243ab50 -> https://github.com/devfile/developer-images/pull/115/commits/48da97d4612123197a573477743c6698523c162b
- https://github.com/redhat-developer/devspaces-images/commit/0bcddb667e023c878ff2432ba17bc3a62c4e1064 -> https://github.com/devfile/developer-images/pull/115/commits/7ac79710c16454500116b51c37e3bdffc30e66de
- https://github.com/redhat-developer/devspaces-images/commit/8eab81da616ce90e6b80ca3d334ef59ccd66dd5c -> https://github.com/devfile/developer-images/pull/115/commits/0b3a58d2143c089d45658b2bd003a47e12ea9a26
- https://github.com/redhat-developer/devspaces-images/commit/9d328a6c8668bebe3d6dfd7f2e6fb17389b36833 -> https://github.com/devfile/developer-images/pull/115/commits/72f53074655fae7f03d06fae4c1b31b909052701
-  https://github.com/redhat-developer/devspaces-images/commit/dbc7d1a0c05fbb462470dda48b738a03a04d1328 -> https://github.com/devfile/developer-images/pull/115/commits/a9669fa1903e1ca37f4b02e871025fd791eb5e56
- https://github.com/redhat-developer/devspaces-images/commit/aabe2265c9a8fa0297191a8f47c09a700f495323 -> https://github.com/devfile/developer-images/pull/115/commits/18b26a72f7ef75ce20889a34d543b3e944729361
- https://github.com/redhat-developer/devspaces-images/commit/07e4857c28e32082458080d6849a0ebe477f80eb -> https://github.com/devfile/developer-images/pull/115/commits/0a20adbf7c02e70f159d592028f299cd350bac5a
- https://github.com/redhat-developer/devspaces-images/commit/f828577293be945ef2ee157f9d1f181e1eff2fbf -> https://github.com/devfile/developer-images/pull/115/commits/d3b7f2e1e0115bac5cc04f5f115e466a826fa1eb

The additional commits required for this change to work on the downstream UDI are as follows:
- [Build stow from sources](https://github.com/redhat-developer/devspaces-images/commit/b99e681f57b75446cbd019e82cb622f75f7051e0)
- [Install vim](https://github.com/redhat-developer/devspaces-images/commit/e96e05312bc3f688bb3fdd1a7fb4d5751f0ab066) 
- [Ensure mountpoint test doesn't cause entrypoint to fail](https://github.com/redhat-developer/devspaces-images/commit/a0af94b282e14c589a73e8d31cbf927524cea50c)

**A note on the repositories used in `content_sets.yml`:** The Code Ready builder **Beta** repo is used for the s390x and ppc64le architectures. The non-beta version of the repo doesn't seem to exist. 

## Testing
This changes touches a lot of tools and configuration files used in the UDI. Thus, it's worth ensuring different workflows still function correctly (e.g. java, mvn, nodejs, go, python). In addition, ensuring podman (with the podman wrapper + kubedock) works correctly is important: see https://github.com/devfile/developer-images/pull/107 for testing notes, though you'll have to modify this [devfile](https://github.com/l0rd/dockerfile-hello-world/blob/kubedock/.devfile.yaml) to use the downstream UDI image built from this PR.  

## Comparison before and after changes

For reference, I'm comparing the UDI built from my changes with `registry.redhat.io/devspaces/udi-rhel8:3.9-18`

#### Before:

```bash
~ $ ls -la
total 16
drwxrwxr-x. 1 root root 184 Oct 31 22:49 .
drwxrwxr-x. 1 root root   8 Oct 31 22:41 ..
drwxrwxr-x. 1 root root  16 Oct 31 22:47 .ansible
-rw-rw-r--. 1 root root  89 Oct 31 22:45 .bashrc
drwxrwxr-x. 1 root root   6 Oct 31 22:45 .cache
drwxrwxr-x. 1 root root  20 Oct 31 22:41 .config
drwxrwxr-x. 1 root root  14 Oct 31 22:45 .java
drwxrwx---. 1 root root  12 Oct 31 22:47 .local
drwxrwxr-x. 1 root root  10 Oct 31 22:49 .npm
drwxrwxr-x. 1 root root  56 Oct 31 22:47 .venv
drwxrwxr-x. 1 root root  36 Oct 31 22:47 che
drwxrwxr-x. 1 root root   6 Oct 31 22:45 go
-rw-rw-r--. 1 root root 509 Oct 31 22:45 group.template
-rw-rw-r--. 1 root root 997 Oct 31 22:45 passwd.template
-rw-rw-r--. 1 root root 558 Oct 31 22:49 xdebug.log
```

#### After:

```bash
~ $ ls -la
total 28
drwxrwx---. 1 user root 242 Nov 13 23:19 .
drwxrwxr-x. 1 root root  22 Nov 13 23:14 ..
drwxrwxr-x. 1 root root  16 Nov 13 23:19 .ansible
-rw-rw-r--. 1 user root  18 Jun 20  2022 .bash_logout
-rw-rw-r--. 1 user root 141 Jun 20  2022 .bash_profile
-rw-rw-r--. 1 user root 376 Jun 20  2022 .bashrc
drwxrwxr-x. 1 root root   6 Nov 13 23:19 .cache
drwxrwxr-x. 1 root root  20 Nov 13 23:19 .config
drwxrwxr-x. 1 root root  14 Nov 13 23:19 .java
drwxrwxr-x. 1 root root  12 Nov 13 23:19 .local
drwxrwxr-x. 1 root root  56 Nov 13 23:19 .venv
-rw-rw----. 1 root root 532 Nov 13 23:19 .viminfo
drwxrwxr-x. 1 root root  36 Nov 13 23:19 che
drwxrwxr-x. 1 root root   6 Nov 13 23:19 go
lrwxrwxrwx. 1 root root  25 Nov 13 23:19 group.template -> ../tooling/group.template
lrwxrwxrwx. 1 root root  26 Nov 13 23:19 passwd.template -> ../tooling/passwd.template
lrwxrwxrwx. 1 root root  21 Nov 13 23:19 xdebug.log -> ../tooling/xdebug.log
```

### `/home/user/` directory tree

Directory tree can be viewed using https://github.com/a8m/tree/tree/master: installed by running `go install github.com/a8m/tree/cmd/tree@latest` then `go clean -cache && go clean -modcache && rm -rf /home/user/.cache/go-build && rm -rf /home/user/.cache/pip`

Unfortunately, the trees are too large to fit within a PR description.

#### Before:

```bash
~ $ cat .bashrc
export PS1='\W `git branch --show-current 2>/dev/null | sed -r -e "s@^(.+)@\(\1\) @"`$ '
```

#### After:

```bash
~ $ cat .bashrc
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
```

Also, here are the 2 new files in `/etc/profile.d/`:

```bash
~ $ cat /etc/profile.d/udi_environment.sh  
alias docker=podman  
~ $ cat /etc/profile.d/udi_prompt.sh
export PS1='\W `git branch --show-current 2>/dev/null | sed -r -e "s@^(.+)@\(\1\) @"`$ '
```

#### Before:

```bash
~ $ echo $PATH | tr : \\n | sort
/bin
/home/user/.java/current/bin
/home/user/.local/bin
/home/user/.npm-global/bin/
/home/user/.venv/bin
/home/user/go/bin
/home/user/node_modules/.bin/
/opt/app-root/src/.npm-global/bin/
/sbin
/usr/bin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/maven/bin
```

#### After:

```bash
~ $ echo $PATH | tr : \\n | sort
/bin
/home/tooling/.java/current/bin
/home/tooling/.local/bin
/home/tooling/.npm-global/bin/
/home/tooling/.venv/bin
/home/tooling/go/bin
/home/tooling/node_modules/.bin/
/home/user/.dotnet/tools
/home/user/.local/bin
/home/user/.local/bin
/home/user/bin
/opt/app-root/src/.npm-global/bin/
/sbin
/usr/bin
/usr/bin
/usr/local/bin
/usr/local/sbin
/usr/sbin
/usr/share/Modules/bin
/usr/share/maven/bin
```


devspaces-udi/Dockerfile
```diff
@@ -193,8 +217,10 @@ RUN \
     ## Rootless podman install #6: rename podman to allow the execution of 'podman run' using
     ##                             kubedock but 'podman build' using podman.orig
     mv /usr/bin/podman /usr/bin/podman.orig && \
+    # Docker alias
+    echo 'alias docker=podman' >> ${PROFILE_EXT} && \
     # set up go/bin folder
-    mkdir /home/user/go/bin -p
+    mkdir /home/tooling/go/bin -p
```
svor, line 223: I know it was added before current changes, but maybe it's redundant to do `mkdir /home/tooling/go/bin -p` 3 times

devspaces-udi/etc/entrypoint.sh
```diff
@@ -83,41 +76,73 @@ if [[ "${KUBEDOCK_ENABLED:-false}" == "true" ]]; then
...
     echo "Kubedock is disabled. It can be enabled with the env variable \"KUBEDOCK_ENABLED=true\""
     echo "set in the workspace Devfile or in a Kubernetes ConfigMap in the developer namespace."
     echo
-    mkdir -p /home/user/.local/bin/
-    ln -f -s /usr/bin/podman.orig /home/user/.local/bin/podman
+    ln -f -s /usr/bin/podman.orig /home/tooling/.local/bin/podman
+fi
+
+#############################################################################
+# Stow: If persistUserHome is enabled, then then the contents of /home/user/
```
svor, line 121:
```suggestion
# Stow: If persistUserHome is enabled, then the contents of /home/user/
```

svor (commented):
@AObuchow do you have a built image that contains current changes, I'd like to test it with existing DS samples

devspaces-udi/Dockerfile
AObuchow: +1 I was thinking of this while working on this. I'm okay to try and remove these redundant mkdir's

AObuchow:
As discussed in standup, I will also include an extra commit to backport https://github.com/devfile/developer-images/pull/135 to the downstream UDI

devspaces-udi/Dockerfile
```diff
@@ -72,15 +84,17 @@ ENV \
     JAVA_HOME_17=/usr/lib/jvm/java-17-openjdk \
     JAVA_HOME_11=/usr/lib/jvm/java-11-openjdk \
     JAVA_HOME_8=/usr/lib/jvm/java-1.8.0-openjdk \
-    JAVA_HOME="/home/user/.java/current" \
-    GOBIN="/home/user/go/bin/" \
-    PATH="/home/user/.local/bin:/home/user/.java/current/bin:/home/user/node_modules/.bin/:/home/user/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/user/go/bin:${PATH:-/bin:/usr/bin}" \
+    JAVA_HOME="/home/tooling/.java/current" \
+    GOBIN="/home/tooling/go/bin/" \
+    PATH="/home/user/.local/bin:/home/tooling/.local/bin:/home/tooling/.java/current/bin:/home/tooling/node_modules/.bin/:/home/tooling/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/tooling/go/bin:${PATH:-/bin:/usr/bin}" \
```
nickboldt, line 89: you have both `user` and `tooling` here... intentional?

devspaces-udi/Dockerfile
```diff
@@ -249,7 +275,7 @@ RUN \
     echo -n "/usr/local/bin/pytest: "; /usr/local/bin/pytest --version && \
     echo -n "/usr/local/bin/yq:     "; /usr/local/bin/yq --version && \
     # set up ~/.venv
-    mkdir -p /home/user; cd /home/user; /usr/bin/python${PYTHON_VERSION} -m venv .venv && \
+    mkdir -p /home/tooling; cd /home/tooling; /usr/bin/python${PYTHON_VERSION} -m venv .venv && \
```
nickboldt, line 278: this dir already exists as of line 199. This isn't bad, just a no-op.

devspaces-udi/content_sets.yml
```diff
@@ -14,13 +14,16 @@ x86_64:
 - rhel-8-for-x86_64-appstream-rpms
 - ocp-tools-4.12-for-rhel-8-x86_64-rpms
 - rhocp-4.12-for-rhel-8-x86_64-rpms
+- codeready-builder-for-rhel-8-x86_64-rpms
 s390x:
 - rhel-8-for-s390x-baseos-rpms
 - rhel-8-for-s390x-appstream-rpms
 - ocp-tools-4.12-for-rhel-8-s390x-rpms
 - rhocp-4.12-for-rhel-8-s390x-rpms
+- codeready-builder-beta-for-rhel-8-s390x-rpms
```
nickboldt, line 23: this being a BETA repo, I hope that doesn't cause problems like turning the UDI container into a `Technology Preview` instead of a GA image.

nickboldt (approved):
probably OK, but some nitpicks noted could be revised.

devspaces-udi/Dockerfile
AObuchow: This was intentional: the line of thought (that [mirrors](https://github.com/devfile/developer-images/pull/115/files#diff-9f21bef4669d474b6ddef409447be53ae91c94cb50b6df9345149a0e89fb5391R101) what is done in the upstream UDI) is that users might want to auto-mount binary's to their workspaces to `/home/user/.local/bin/` (it's much more natural to mount binaries to `~/.local/bin/ `rather than `/home/tooling/.local/bin/`)

devspaces-udi/Dockerfile
AObuchow: Good catch, IMO this can be safely removed.

devspaces-udi/content_sets.yml
AObuchow: That's my worry as well 🫠 IIRC, the codeready repo was required for the [`texinfo`](https://access.redhat.com/downloads/content/texinfo/6.7-15.el9/x86_64/fd431d51/package) and [`texinfo-tex`](https://access.redhat.com/downloads/content/texinfo-tex/x86_64/package-latest) packages, needed to build stow. We _might_ be able to do without these packages using some flags in the build process of stow, to disable generation of the documentation. Unfortunately, I haven't been able to figure this out yet.

devspaces-udi/Dockerfile
```diff
@@ -107,8 +121,9 @@ LABEL summary="$SUMMARY" \
 ADD etc/storage.conf $HOME/.config/containers/storage.conf
 ADD etc/entrypoint.sh /entrypoint.sh
 COPY $REMOTE_SOURCES $REMOTE_SOURCES_DIR
-RUN echo 'alias docker=podman' >> /home/user/.bashrc
 COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
+COPY --chown=0:0 etc/.stow-local-ignore /home/tooling/
+RUN mkdir -p /home/tooling/
```
amisevsk, line 125: Are these in the wrong order? it looks like we're copying to a directory and then creating it.

devspaces-udi/content_sets.yml
amisevsk: It might actually block shipping the image later in general -- if you check those URLs, texinfo and texinfo-tex are only available on x86_64 and aarch64

Might be safer to only do the stow stuff if the arch is supported

devspaces-udi/Dockerfile
```diff
@@ -158,8 +173,15 @@ RUN \
     mkdir -p /opt && \
     # add user and configure it
     useradd -u 1000 -G wheel,root -d /home/user --shell /bin/bash -m user && \
+    # Bash-related files are backed up to /home/tooling/ incase they are deleted when persistUserHome is enabled.
+    cp /home/user/.bashrc /home/tooling/.bashrc && \
+    cp /home/user/.bash_profile /home/tooling/.bash_profile && \
+    chown 10001 /home/tooling/.bashrc /home/tooling/.bash_profile && \
+    # $PROFILE_EXT contains all additions made to the bash environment
+    touch ${PROFILE_EXT} && chown 10001 ${PROFILE_EXT} && \
```
amisevsk, line 180: Why `chown 10001`?

devspaces-udi/etc/entrypoint.sh
```diff
@@ -83,41 +76,73 @@ if [[ "${KUBEDOCK_ENABLED:-false}" == "true" ]]; then
...
+      echo "Replacing podman with podman-wrapper.sh..."
+
+      ln -f -s /usr/bin/podman-wrapper.sh /home/tooling/.local/bin/podman
+
+      export TESTCONTAINERS_RYUK_DISABLED="true"
+      export TESTCONTAINERS_CHECKS_DISABLE="true"
+
+      echo "Done."
+      echo
+    else
```
amisevsk, line 108: Nit: indentation here seems to switch between 2 and 4 spaces.

devspaces-udi/etc/entrypoint.sh
```diff
@@ -83,41 +76,73 @@ if [[ "${KUBEDOCK_ENABLED:-false}" == "true" ]]; then
...
     echo "set in the workspace Devfile or in a Kubernetes ConfigMap in the developer namespace."
     echo
-    mkdir -p /home/user/.local/bin/
-    ln -f -s /usr/bin/podman.orig /home/user/.local/bin/podman
+    ln -f -s /usr/bin/podman.orig /home/tooling/.local/bin/podman
+fi
+
+#############################################################################
+# Stow: If persistUserHome is enabled, then the contents of /home/user/
+# will be mounted by a PVC and overwriten. In this case, we use stow to
```
amisevsk, line 122:
```suggestion
# will be mounted by a PVC and overwritten. In this case, we use stow to
```

AObuchow:
I pushed additional commits based on the feedback so far. @svor @nickboldt thank you both for the reviews.

https://github.com/redhat-developer/devspaces-images/pull/525/commits/b8ecb9a31fe64b985490dc6c25c1cbacb4b3f2b0 is a port of https://github.com/devfile/developer-images/pull/135, with Gradle and SBT being excluded since they are not being installed on the downstream UDI.

The newest changes are available for testing at: `registry-proxy.engineering.redhat.com/rh-osbs/devspaces-udi-rhel8:devspaces-3-rhel-8-containers-candidate-68595-20231114193307-x86_64`

devspaces-udi/etc/entrypoint.sh
```diff
@@ -83,41 +76,73 @@ if [[ "${KUBEDOCK_ENABLED:-false}" == "true" ]]; then
...
+
+if [ $HOME_USER_MOUNTED -eq 0 ] && [ ! -f $STOW_COMPLETE ]; then
+    # Create symbolic links from /home/tooling/ -> /home/user/
+    stow . -t /home/user/ -d /home/tooling/ --no-folding -v 2 > /tmp/stow.log 2>&1
+    # Vim does not permit .viminfo to be a symbolic link for security reasons, so manually copy it
+    cp /home/tooling/.viminfo /home/user/.viminfo
+    # We have to restore bash-related files back onto /home/user/ (since they will have been overwritten by the PVC)
+    # but we don't want them to be symbolic links (so that they persist on the PVC)
+    cp /home/tooling/.bashrc /home/user/.bashrc
+    cp /home/tooling/.bash_profile /home/user/.bash_profile
```
amisevsk, line 144: I think we want `-n / --no-clobber` here -- this risks overwriting these files if they already exist in `/home/user`.

Also, does this still work if I e.g. mount a `.bashrc` from a configmap?

devspaces-udi/content_sets.yml
amisevsk: Maybe a dumb question, but does install succeed even without those packages? It looks like we're not copying them even if built.

devspaces-udi/Dockerfile
AObuchow: +1, they should be re-ordered for clarity.
Surprisingly, `/home/tooling/.stow-local-ignore` is present in the image. I guess the COPY directive also creates `/home/tooling/` while it's at it.

devspaces-udi/Dockerfile
AObuchow: This is a leftover from the upstream UDI image. In the downstream UDI image, it doesn't seem like we're creating a user (10001) -- IIRC, the user is pre-created by OpenShift.

Will remove the `chown`'s, good catch.

devspaces-udi/etc/entrypoint.sh
AObuchow: Thank you, the diff was bugging me, but I wasn't sure why :P Will fix that.

devspaces-udi/etc/entrypoint.sh
AObuchow: +1 on using `--no-clobber` (that also should be added to the upstream UDI).

I'll verify that `.bashrc `configmap case.

AObuchow:
> It might actually block shipping the image later in general -- if you check those URLs, texinfo and texinfo-tex are only available on x86_64 and aarch64

Good catch @amisevsk. I'm not sure why I was under the impression texinfo and texinfo-tex were available in the Code Ready Beta repo's. 

> Might be safer to only do the stow stuff if the arch is supported

Not sure yet what the cleanest and most feasible way to do this is: do we create a second dockerfile for PPC & s390x? Do we do a runtime check on the architecture? 

It's probably worth investigating if there's a simple way to disable the need for `tex` when [building stow](https://github.com/aspiers/stow/blob/master/Makefile.am#L180-L253). Will look into that further.

nickboldt:
> cleanest and most feasible way to do this is: do we create a second dockerfile for PPC & s390x? 

100% no, that won't work

> Do we do a runtime check on the architecture?

Yes, we do this when support on various arches varies, like the versions of dotnet available on Z & P: 

https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile#L284-L290

So you would want to ONLY include stow on x64, unless you can figure out a way to build it on Z & P too. 

You can ask for help on those arches from @Sachin-Itagi or @ghatwala, if you want to just implement this for x64 and ignore the other two arches for now. 

Either way, we'll likely need to loop in @deerskindoll to add a new RNaKI item if the support is different by arch.

svor:
@AObuchow it's not possible to build a java project with current changes, when executing **mvn clean install**. I didn't investigate what is the reason, but I see an error:
```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.092 s
[INFO] Finished at: 2023-11-15T13:27:58Z
[INFO] ------------------------------------------------------------------------
[ERROR] Plugin org.apache.maven.plugins:maven-clean-plugin:2.5 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-clean-plugin:jar:2.5: Could not transfer artifact org.apache.maven.plugins:maven-clean-plugin:pom:2.5 from/to central (https://repo.maven.apache.org/maven2): No connector factories available -> [Help 1]
``` 
You can try it with lombok DS sample: [Contribute](https://devspaces.apps.sandbox-stage.gb17.p1.openshiftapps.com/dashboard/#/https://github.com/svor/lombok-project-sample/tree/udi-pr-525)

l0rd (approved):
I have did some tests building, running containers and switching java version and everything worked just fine. LGTM :thumbsup:

AObuchow:
@svor Thank you for the feedback, I'll be working on reproducing and trying to fix the issues with maven & dotnet that you described.

AObuchow:
@l0rd Glad to hear things seemed to work correctly in your testing.

AObuchow:
> > Do we do a runtime check on the architecture?
> 
> Yes, we do this when support on various arches varies, like the versions of dotnet available on Z & P:
> 
> [`devspaces-3-rhel-8`/devspaces-udi/Dockerfile#L284-L290](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile?rgh-link-date=2023-11-15T02%3A58%3A50Z#L284-L290)
> 
> So you would want to ONLY include stow on x64, unless you can figure out a way to build it on Z & P too.
> 

Thank you for the guidance @nickboldt.

Something strange to note: my brew builds have actually been succeeding (including s390x & PPC). Looking at the [build logs,](http://download.eng.bos.redhat.com/brewroot/work/tasks/2004/56952004/) I can see that stow seems to be getting built, and its version is being outputted correctly after building: `stow (GNU Stow) version 2.3.2`.

In the logs I see that there is a texinfo-tex package for both architectures (`texinfo-tex-6.5-7.el8.ppc64le` & `texinfo-tex-6.5-7.el8.s390x`) and a texinfo package (`texinfo-6.5-7.el8.ppc64le` & `texinfo-6.5-7.el8.s390x`).

This doesn't mean we're in the clear yet as we're still using the Code Ready Beta repo's however.

AObuchow:
@svor @nickboldt For dotnet, it seems that the install is failing:

```
2023-11-14 19:49:31,920 - atomic_reactor.tasks.binary_container_build - INFO - Error:
2023-11-14 19:49:31,920 - atomic_reactor.tasks.binary_container_build - INFO -  Problem: cannot install the best candidate for the job
2023-11-14 19:49:31,920 - atomic_reactor.tasks.binary_container_build - INFO -   - nothing provides netstandard-targeting-pack-2.1(x86-64) >= 7.0.113-2.el8_9 needed by dotnet-sdk-7.0-7.0.113-2.el8_9.x86_64 from rhel-8-for-x86_64-appstream-rpms
```

I'm not sure yet why this is happening. Adding an extra repository shouldn't cause this issue? Unless a subset of alternative packages were mistakenly installed from the additional Code Ready builder repo, and their matching dependencies (required for dotnet) aren't present. Maybe I need to disable the code ready repo after stow has been built?

devspaces-udi/content_sets.yml
AObuchow: > does install succeed even without those packages? It looks like we're not copying them even if built.

If you're referring to the stow build: it fails to build without the `texinfo` and `texinfo-tex` packages. Though, I might be able to disable documentation generation during stow's build to remove the need for these packages.

devspaces-udi/etc/entrypoint.sh
AObuchow: > Also, does this still work if I e.g. mount a .bashrc from a configmap?

I tested this and surprisingly, my `~/.bash_profile` (mounted with an automount configmap) was there instead of the one from `/home/tooling/`.

Still, using `--no-clobber` is preferable.

AObuchow:
@svor `mvn` and `dotnet` should now be fixed.

For maven, the wrong maven-resolver package was accidentally getting pulled from the code ready builder repo so I disabled it after building stow (though, maybe this should happen earlier in the dockerfile.. it looks kind of ugly, but I wasn't sure about introducing an extra layer).

For dotnet, I think there was some weird upgrade that was in-flight for dotnet's dependencies when I built the image. It seems to work now.

@amisevsk I addressed your feedback. I realized certain changes belonged to the wrong commits, which is why you see me deleting and re-adding the same lines (with slight modifications based on your feedback).

AObuchow:
Something that came to mind: maven is using `/root/.m2` as the local repository location to download artifacts. Ideally it should be `/home/user/.m2`, but I haven't found the right environment variable to modify this behaviour, and the default `settings.xml` mentions that the default local maven repository should be `$HOME/.m2` (but that is not the case ?)

Also, we might want to change GOBIN and GOPATH to point to `/home/user/go/` instead of `/home/tooling/go` so that go modules that are downloaded persist on the PVC if persistUserHome is enabled. I seem to have missed this detail when making the upstream PR for this change.

AObuchow:
The latest changes can be tested with `registry-proxy.engineering.redhat.com/rh-osbs/devspaces-udi-rhel8:devspaces-3-rhel-8-containers-candidate-68148-20231116060632-x86_64`

devspaces-udi/Dockerfile
```diff
@@ -193,12 +224,21 @@ RUN \
...
-    # set up go/bin folder
-    mkdir /home/user/go/bin -p
+    # Docker alias
+    echo 'alias docker=podman' >> ${PROFILE_EXT}
 
 RUN \
     # configure runtimes
     ########################################################################
+    # Common tooling configuration directories
+    ########################################################################
```
AObuchow, line 234: extra newline should be removed

svor (approved):
tested with registry-proxy.engineering.redhat.com/rh-osbs/devspaces-udi-rhel8:devspaces-3-rhel-8-containers-candidate-68148-20231116060632-x86_64

everything works fine!

AObuchow:
@svor Thank you for the review!

Assuming nothing critical comes up during the rest of the review, my last challenge for this PR is to see if I can get stow to build without requiring the texinfo and texinfo-tex packages to remove the dependency on the code ready builder repo.

AObuchow:
> For maven, the wrong maven-resolver package was accidentally getting pulled from the code ready builder repo so I disabled it after building stow (though, maybe this should happen earlier in the dockerfile.. it looks kind of ugly, but I wasn't sure about introducing an extra layer).

@nickboldt please let me know when you get a chance if you think the placement of the "disabling code ready repo's" code should be moved elsewhere :)

AObuchow:
@amisevsk Once you verify my fix up commits I'll rebase & squash this PR, and it should be good to merge. Currently, doing one last build after rebasing.

AObuchow:
Unfortunately, now the s390x and PPC builds are [complaining](https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=57030588) that they don't have the texinfo and texinfo-tex packages available, despite [previous builds](https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=57030588) having them just 2 days ago. No idea what changed 🫠.

I will have to disable stow on these platforms for now, then this PR should be finally good to merge.

AObuchow:
I added 2 commits, which should contain the final required changes:
- https://github.com/redhat-developer/devspaces-images/pull/525/commits/e0e7f184a1df9b3c650876935e0bfe0ce23f0299 fixes the upgrade to python 3.11
- https://github.com/redhat-developer/devspaces-images/pull/525/commits/58b9cec1244fe6fdca008318b8b3eee4cfea4a9d addresses the issue with the texinfo and  texinfo-tex packages not being found. They seemed to have moved to the official Code Ready Builder repo's for 390x & ppc \o/

AObuchow:
The latest [build](https://pkgs.devel.redhat.com/cgit/containers/devspaces-udi/commit/?h=private-aobuchow-persistUserHome&id=62150701eac7ef973785ffd9c9c73b196d2d9422) can be tested with `registry-proxy.engineering.redhat.com/rh-osbs/devspaces-udi-rhel8:devspaces-3-rhel-8-containers-candidate-93078-20231117182211-x86_64`

AObuchow:
Squashed all the fixup commits, this PR should now be good to finally merge.

amisevsk:
What a strangely fortunate change :D

devspaces-udi/Dockerfile
```diff
@@ -55,13 +54,25 @@ RUN \
...
+    # END stow
+
 # https://registry.access.redhat.com/ubi8-minimal
 FROM ubi8-minimal:8.8-1072
 
 USER root
 
 ENV \
-    HOME=/home/user \
+    # We install everything to /home/tooling/ as /home/user/ may get overriden, see github.com/eclipse/che/issues/22412
```
amisevsk, line 74:
```suggestion
    # We install everything to /home/tooling/ as /home/user/ may get overridden, see github.com/eclipse/che/issues/22412
```

devspaces-udi/Dockerfile
```diff
@@ -72,15 +83,17 @@ ENV \
     JAVA_HOME_17=/usr/lib/jvm/java-17-openjdk \
     JAVA_HOME_11=/usr/lib/jvm/java-11-openjdk \
     JAVA_HOME_8=/usr/lib/jvm/java-1.8.0-openjdk \
-    JAVA_HOME="/home/user/.java/current" \
-    GOBIN="/home/user/go/bin/" \
-    PATH="/home/user/.local/bin:/home/user/.java/current/bin:/home/user/node_modules/.bin/:/home/user/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/user/go/bin:${PATH:-/bin:/usr/bin}" \
+    JAVA_HOME="/home/tooling/.java/current" \
+    GOBIN="/home/tooling/go/bin/" \
+    PATH="/home/user/.local/bin:/home/tooling/.local/bin:/home/tooling/.java/current/bin:/home/tooling/node_modules/.bin/:/home/tooling/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/tooling/go/bin:${PATH:-/bin:/usr/bin}" \
```
amisevsk, line 88: We need to be very careful with paths and env vars here, and I'm not sure updating these to use `/home/tooling` is the right way to go. Some examples:

* Setting `GOBIN` to `/home/tooling` means certain things would be broken -- e.g. if I install goimports (`go install golang.org/x/tools/cmd/goimports@latest`), this change is lost on restart
* Setting `JAVA_HOME` means (I believe) that any changes to java version are lost on container restart.

Further, I'm worried that setting `PATH` to search `/home/tooling` instead of `/home/user` will lead to strange behavior. If I `npm install` something, will it go in `/home/tooling/.npm-global/bin/` or `/home/user/.npm-global/bin/`? If it goes in the former, it will disappear on container restart, and if it goes in the latter it will not be on my path.

I think the existence of `/home/tooling` should be invisible to the user unless they dig into symlink destinations. We should be fine to leave these env vars pointing at `/home/user` as path resolution _should_ still find any binaries.

devspaces-udi/Dockerfile
```diff
@@ -104,11 +117,13 @@ LABEL summary="$SUMMARY" \
       io.openshift.expose-services="" \
       usage=""
 
+RUN mkdir -p /home/tooling/
```
amisevsk, line 120: Still a no-op, covered by line 26 :stuck_out_tongue: 

Maybe move this to the very top to avoid this continuing cat-and-mouse game?

devspaces-udi/Dockerfile
```diff
@@ -104,11 +117,13 @@ LABEL summary="$SUMMARY" \
...
 ADD etc/storage.conf $HOME/.config/containers/storage.conf
 ADD etc/entrypoint.sh /entrypoint.sh
 COPY $REMOTE_SOURCES $REMOTE_SOURCES_DIR
-RUN echo 'alias docker=podman' >> /home/user/.bashrc
 COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
+RUN mkdir -p /home/tooling/
```
amisevsk, line 125: See above :)

devspaces-udi/Dockerfile
```diff
@@ -121,7 +136,16 @@ COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
...
-    dnf -y -q module reset   container-tools       maven     nodejs                 python3.11                   php; \
+    # Disable code ready builder repo's to prevent accidentally installing incorrect packages
+    dnf -y -q install 'dnf-command(config-manager)' && \
+    if [[ "$(uname -m)" == 'x86_64' ]]; then \
+        dnf config-manager --set-disabled codeready-builder-for-rhel-8-x86_64-rpms ; \
+    elif [[ "$(uname -m)" == 's390x' ]]; then \
+        dnf config-manager --set-disabled codeready-builder-for-rhel-8-s390x-rpms ; \
+    elif [[ "$(uname -m)" == 'ppc64le' ]]; then \
+        dnf config-manager --set-disabled codeready-builder-for-rhel-8-ppc64le-rpms ; \
+    fi && \
```
amisevsk, line 147: I think you can get away with just using `dnf --disablerepo <name> (...)` instead of installing `config-manager` (also accepts glob patterns so you should be able to do `dnf --disablerepo codeready-*` to avoid branching on arch).

Alternatively, is there a way to set priority on repos in `content-sets.yml`?

devspaces-udi/Dockerfile
```diff
@@ -158,8 +182,14 @@ RUN \
     mkdir -p /opt && \
     # add user and configure it
     useradd -u 1000 -G wheel,root -d /home/user --shell /bin/bash -m user && \
+    # Bash-related files are backed up to /home/tooling/ incase they are deleted when persistUserHome is enabled.
```
amisevsk, line 185:
```suggestion
    # Bash-related files are backed up to /home/tooling/ in case they are deleted when persistUserHome is enabled.
```

devspaces-udi/etc/entrypoint.sh
amisevsk: Still looks weird -- indentation of lines 89-->107 should match 109-110

amisevsk (commented):
I have a few comments -- I'm worried about edge cases around setting up paths and env vars (hard to actually verify all potential edge cases for all installed tools, though).

devspaces-udi/Dockerfile
SDawley: Overwritten actually I think :smile:

devspaces-udi/Dockerfile
```diff
@@ -72,15 +83,17 @@ ENV \
...
+    JAVA_HOME="/home/tooling/.java/current" \
+    GOBIN="/home/tooling/go/bin/" \
+    PATH="/home/user/.local/bin:/home/tooling/.local/bin:/home/tooling/.java/current/bin:/home/tooling/node_modules/.bin/:/home/tooling/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/tooling/go/bin:${PATH:-/bin:/usr/bin}" \
     MANPATH="/usr/share/man:${MANPATH}" \
     JAVACONFDIRS="/etc/java${JAVACONFDIRS:+:}${JAVACONFDIRS:-}" \
     XDG_CONFIG_DIRS="/etc/xdg:${XDG_CONFIG_DIRS:-/etc/xdg}" \
     XDG_DATA_DIRS="/usr/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}" \
     M2_HOME="/usr/share/maven" \
     PKG_CONFIG_PATH="/usr/lib64/pkgconfig${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}}" \
+    KUBECONFIG=/home/user/.kube/config \
```
SDawley, line 95: Just double checking, is this a file that will get overwritten and should go in /home/tooling?

devspaces-udi/container.yaml
```diff
@@ -64,3 +64,10 @@ remote_sources:
     ref: 9d21955b52e4905d916d24e724dcad195aef3515
     pkg_managers:
       - gomod
+
+- name: stow
+  remote_source:
+    repo: https://github.com/aspiers/stow
+    ref: 4ef5eca4a9d107b24e712bb4c2c91f47e7e0fb85
```
SDawley, line 71: How often will stow need to be updated?
We update the ref for images during sync-to-downstream, but ideally this isn't something we'll update manually (often).

devspaces-udi/Dockerfile
nickboldt: you could also 
`for d in /etc/yum.repos.d/*codeready-builder*; do sed -i $d -r -e "s/enabled=1/enabled=0/"; done`
do disable the repos in a single line. 

Or even simpler:
`find /etc/yum.repos.d/ -type f -name "*codeready-builder*" -delete`

devspaces-udi/Dockerfile
AObuchow: Good catch :) 

IIRC the `~/.kube/config` file gets injected into the workspace after startup (after the PVC has  been mounted and overwritten` /home/user/`) _but_ the expected default value of `$KUBECONFIG` is supposed to be `$HOME/.kube/config`. `$HOME` is set to `/home/user/` but I figured it'd be best to explicitly set this environment variable since it's value is now referenced in the entrypoint: https://github.com/redhat-developer/devspaces-images/pull/525/files#diff-f507e1fba3158ecb25e3f6c3e8abe0bbf3b97f9c208d9239c6a00742e3334daaR88.

I also think that setting `KUBECONFIG` to` /home/tooling/.kube/config` did not work when attempted in the upstream UDI.

devspaces-udi/container.yaml
AObuchow: > How often will stow need to be updated? 

Ideally, never. I think it would realistically only need to be updated if new security vulnerabilities come out for it or its dependencies (though the dependency CVE's should be covered by the Red Hat repo's we're pulling them from).

devspaces-udi/Dockerfile
AObuchow: I think an approach that doesn't depend on the arch would be _much_ cleaner than what I have now. All the solutions proposed by @amisevsk  and @nickboldt sound good -- I will make some changes there.

> Alternatively, is there a way to set priority on repos in content-sets.yml?

Not that I'm aware of :(

devspaces-udi/Dockerfile
AObuchow: Oops, good catch. I wasn't sure if the mkdir in question is a no-op by line 26 since they occur in different container images (go-toolest vs. UBI)

devspaces-udi/Dockerfile
AObuchow: This is a really good point & something I had concerns with :+1: 

# $PATH:

In the upstream UDI, we set everything to point to `/home/tooling/` instead of `/home/user/` for the following 2 reasons:
1. sdkman breaks if `$SDKMAN_DIR` is in a directory that contains links (e.g. `/home/user/`) instead of actual files. This is not relevant in the downstream UDI since sdkman is not used.
2. The [Kubernetes docs](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion) mention that there is no guarantee whether the containers entrypoint or postStart commands will run first on container creation. Thus, it's possible to encounter a race condition if persistUserHome is enabled and a postStart command tries executing a binary that resides in `/home/user/...` before stow has run in the [entrypoint](https://github.com/redhat-developer/devspaces-images/pull/525/files#diff-f507e1fba3158ecb25e3f6c3e8abe0bbf3b97f9c208d9239c6a00742e3334daaR138).

I think a best-effort approach would be to include both `/home/user/...` and `/home/tooling/...` (afterwards) in $PATH. Though this is awkward, it would result in first prioritizing binaries and other files (e.g. npm resources) that are persisting in `/home/user/`, and then falling back to `/home/tooling/` if the entrypoint hasn't yet run (for postStart commands). 

# Java

I just tested, and manually switching java version's is actually currently broken: the symbolic links in `/home/user/.java/current/` get updated, but since PATH points to `/home/tooling/.java/current/`, the old version of java found in `/home/tooling/.java/current/` gets used instead. So $PATH's entry for Java to needs to point to `/home/user/...`. Or the [entrypoint](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/etc/entrypoint.sh#L63) needs to explicitly create symbolic links from `/home/tooling/.java/current/` instead of `$HOME/.java/current/`.

I'm actually uncertain where `JAVA_HOME` should point to: `/home/user/` would be best, but I'm not sure if JAVA_HOME could potentially be needed in a postStart command? On my local system, `$JAVA_HOME` is unset, so I'm hopeful we could set it to `/home/user/.java/current/`

# Go Bin

> Setting GOBIN to `/home/tooling` means certain things would be broken -- e.g. if I install goimports (`go install golang.org/x/tools/cmd/goimports@latest`), this change is lost on restart

This is correct, and my concern as well. 

Doing a `go help go path` states that: (where DIR is a directory listed in the GOPATH):
```
If the GOBIN environment variable is set, commands are installed to the directory it names instead of DIR/bin. GOBIN must be an absolute path.
```

GOBIN affects where Go binaries are _installed_, but $PATH determines where Go binaries are located when trying to execute them.

I believe it's safe to set GOBIN to `/home/user/` since we know the PVC will be mounted. PATH should include both `/home/user/.go/bin/` (for executing user-installed Go binaries) and `/home/tooling/.go/bin/` for running gopls and (more importantly) kubedock as postStart commands or in the [entrypoint](https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/etc/entrypoint.sh#L100).

AObuchow: To summarize my above (lengthy) comment:
- `$PATH` should include both `/home/user/...` and `/home/tooling/...`, with the `/home/user/` entries occurring first for higher priority, and `/home/tooling/` entries occurring later so that postStart commands can always execute any binaries that reside within that directory.
- `$JAVA_HOME` should (probably?) be set to `/home/user/.java/current/`
- `$GOBIN` should be set to `/home/user/.go/bin/`

devspaces-udi/Dockerfile
AObuchow: I verified that the mkdir at line 26 (in the first stage of the build) won't affect the second stage of the build.

To test this, I created the following dockerfile and saw that it failed to build:

```Dockerfile
FROM fedora:37

USER 0
RUN mkdir -p /home/tooling/

FROM fedora:38

USER 0
RUN stat /home/tooling/
```


```
=> ERROR [stage-1 2/2] RUN stat /home/tooling/                                                                                                                       0.7s
------
 > [stage-1 2/2] RUN stat /home/tooling/:
0.653 stat: cannot statx '/home/tooling/': No such file or directory
```

amisevsk: Ah I didn't realize it was a different builder image :+1:

AObuchow:
@amisevsk I added 2 extra commits to address the PATH issue and code-ready repo disable. Though I am still waiting on my brew build to finish.

AObuchow:
My fix for the PATH problem was incorrect... sorry about that. Fixing it.

devspaces-udi/Dockerfile
```diff
@@ -83,9 +83,11 @@ ENV \
...
     JAVA_HOME_11=/usr/lib/jvm/java-11-openjdk \
     JAVA_HOME_8=/usr/lib/jvm/java-1.8.0-openjdk \
-    JAVA_HOME="/home/tooling/.java/current" \
-    GOBIN="/home/tooling/go/bin/" \
-    PATH="/home/user/.local/bin:/home/tooling/.local/bin:/home/tooling/.java/current/bin:/home/tooling/node_modules/.bin/:/home/tooling/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/tooling/go/bin:${PATH:-/bin:/usr/bin}" \
+    JAVA_HOME="/home/user/.java/current" \
+    GOBIN="/home/user/go/bin/" \
+    PATH="/home/user/.local/bin:/home/user/.java/current/bin:/home/user/node_modules/.bin/:/home/user/.npm-global/bin/:/opt/app-root/src/.npm-global/bin/:/usr/share/maven/bin:/usr/bin:/home/user/go/bin:${PATH:-/bin:/usr/bin}" \
+    # We have $PATH entries in /home/tooling/ as well to ensure binaries can be executed in case /home/user/ has been ovewritten and stow has not yet run in the entrypoint
+    PATH="${PATH}:/home/tooling/.local/bin:/home/tooling/.java/current/bin:/home/tooling/node_modules/.bin/:/home/tooling/.npm-global/bin/:/home/tooling/go/bin" \
```
AObuchow, line 90: Unfortunately, the second PATH environment variable declaration overwrites the first one at line 88, so this won't work.

The alternatives are:
- Use a second ENV directive immediately afterwards to append /home/tooling/ entries to path (which makes it kind of hidden from the first ENV directive that modifies PATH 
- Do everything in one ENV directive: messy, but more explicit and discoverable to the reader

AObuchow: I went with the second approach though am open to the first approach.

AObuchow:
The latest changes can be tested with: `docker run -ti --rm    registry-proxy.engineering.redhat.com/rh-osbs/devspaces-udi-rhel8:devspaces-3-rhel-8-containers-candidate-43055-20231120202203-x86_64 bash`

AObuchow:
@nickboldt If this gets a +1 from @amisevsk & @SDawley, I will squash the remaining fixup commits and this should be good to finally merge 🥳

devspaces-udi/Dockerfile
```diff
@@ -121,7 +137,11 @@ COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
 RUN \
     # install all the rpms and modules
     microdnf install -y dnf && \
-    dnf -y -q module reset   container-tools       maven     nodejs                 python3.11                   php; \
+    # Disable code ready builder repo's to prevent accidentally installing incorrect packages
```
nickboldt, line 140:
```suggestion
    # Disable codeready-builder repos to prevent accidentally installing incorrect packages
```

devspaces-udi/Dockerfile
```diff
@@ -121,7 +137,11 @@ COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
 RUN \
     # install all the rpms and modules
     microdnf install -y dnf && \
-    dnf -y -q module reset   container-tools       maven     nodejs                 python3.11                   php; \
+    # Disable codeready-builder repos to prevent accidentally installing incorrect packages
+    dnf -y -q install 'dnf-command(config-manager)' && \
+    dnf config-manager --set-disabled codeready-* && \
+    fi && \
```
nickboldt, line 143:
```suggestion
    \
```

devspaces-udi/Dockerfile
```diff
@@ -121,7 +137,11 @@ COPY --chown=0:0 etc/podman-wrapper.sh /usr/bin/
...
+    dnf -y -q module reset   container-tools       maven     nodejs                 php; \
```
nickboldt, line 144:
```suggestion
    dnf -y -q module reset container-tools maven nodejs php; \
```

devspaces-udi/Dockerfile
nickboldt:
```
dnf config-manager --set-disabled codeready-*
```

Seems nice and simple.

devspaces-udi/Dockerfile
AObuchow: I removed the line, good catch :+1:

amisevsk (approved):

PR #532, opened 2023-12-01
fix: fixes for Cachito built-in sync

mkuznyetsov:
- fix initialization of script dir variables (would fail on non-3-x branch execution otherwise
- read vscode-ripgrep version straight from yarn.lock, instead of package.json (don't want to deal with possible ranges in package.json)
- edit script output


nickboldt (approved):
seems reasonable.

However before merging it would be nice to have more description for the reasons for this change.

`No description provided.` and `fixes` are NOT sufficient.

PR #542, opened 2024-01-09
fix: get correct tag of openvsx-server repo

svor:
Get tag of che-openvsx repository during the build to pull correct version

Related issue: https://issues.redhat.com/browse/CRW-5454


SDawley (approved):
Other than the fact that the new code block isn't aligned, looks good to me.

svor:
> Other than the fact that the new code block isn't aligned, looks good to me.

thanks, I've fixed the alignment

PR #556, opened 2024-03-14
use cachito to build and prepare ovsx 0.8.3 in plugin registry part

svor:
Related issue: https://issues.redhat.com/browse/CRW-3336

- Add ovsx.Dockerfile that should replace old way to install ovsx npm library to get the source files via cachito 
- The **package.json** file with generated **yarn.lock** for ovsx npm package are located at **devspaces-pluginregistry/cachito/ovsx**
- The **container.yaml** was updated by adding yarn package manager to build ovsx library. (`ref: 03588209388ce14179c03e3a9833d5c157b3b846` should be updated after merging this PR)
- I've also updated **synch.s** file to don't break upstream/local build. With a new version of **sync.sh** the final Dockerfile will be updated by adding ovsx.Dockerfile at the beginning and old part to injecting ovsx will be also replaced.

The test branch on pkgs.devel is https://pkgs.devel.redhat.com/cgit/containers/devspaces-pluginregistry/log/?h=private-vsvydenk-ovsx

The last successful build is https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=59712136


SDawley:
@svor Just curious, if ovsx is going to remain it's own dockerfile how is it going to be pulled into the pluginregistry cachito build?

svor:
> @svor Just curious, if ovsx is going to remain it's own dockerfile how is it going to be pulled into the pluginregistry cachito build?

@SDawley sorry I didn't push changes related to the Dockerfile that is used by brew. This Dockerfile is generated by executing sync.sh script.

ovsx.Dockerfile is not used separately, it should be a part of Dockerfile that is used by brew, the sync.sh script adds its content to the main Dockerfile. My idea having it in a separate file to don't overload sync.sh script

SDawley (approved):
OK that makes sense. Approved. 
Once this is merged I can add in my changes to replace root-local.tgz and resources.tgz, and pull in the PNC artifacts. 

Then all that should be left is getting the plugins to build consistently and use those.

PR #570, opened 2024-05-10
chre: update ansible image in allowed list 

svor:
Since ansible image was updated in the sample (PR with changes: https://github.com/devspaces-samples/ansible-devspaces-demo/pull/28) we need to update it in the script that validate images

Related issue: https://issues.redhat.com/browse/CRW-6290


SDawley (approved):
lgtm

PR #577, opened 2024-05-22
fix: fix build of devspaces-code

RomanNikitenko:
Issue: https://issues.redhat.com/browse/CRW-6313

rh-pre-commit.version: 2.2.0


RomanNikitenko:
I don't think there is a downside:
- it's a known issue on the Node.js side, you can read more about it here: https://github.com/nodejs/node/issues/52229.
- in my PR I just applied their proposal from the Node.js team.

PR #547, opened 2024-01-29
feat: Support ubi9-based containers for running VS Code editor

RomanNikitenko:
### What does this PR do?

- At the moment root `Dockerfile` for the local build of the `devspaces-code` doesn't work, so it's not possible to build `devspaces-code` assembly. The current PR fixes that `Dockerfile`.
- Also the PR adds ability to build `ubi9`-based assembly for both: local build and build on Brew.
- The PR does **not** add `musl/alpine-based` containers support for running `devspaces-code`

#### Support ubi9-based containers for running VS Code editor:
- Node.js provided by `registry.access.redhat.com/ubi9/nodejs-18` is used for running VS Code in a `ubi9-based` container
- `libbrotli` library from the same image is copied to a user's container as a dependency for Node.js
- `LD_LIBRARY_PATH` env variable to define additional shared libs place (`libbrotli`) 

### What issues does this PR fix?

https://issues.redhat.com/browse/CRW-5629

### How to test this PR?
**Image for testing:** quay.io/rnikitenko/devspaces-code:ubi8-ubi9-jenkins

**Successful build on jenkins**: [DS_CI/job/get-sources-rhpkg-container-build_3.x/6138](https://main-jenkins-csb-crwqe.apps.ocp-c1.prod.psi.redhat.com/job/DS_CI/job/get-sources-rhpkg-container-build_3.x/6138/consoleFull)

I tested the following images as a container for running Devspaces Code:

| Image  | Workspace starts | Devfile registry link | Link to start a workspace |
| :-------------: | :-------------: | :-------------: | :-------------: |
| `registry.access.redhat.com/ubi8`  | + <br /> Git not found | - | [click here](https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com/#https://github.com/RomanNikitenko/test-redhat-nodejs/tree/ubi8-devspaces-jenkins)  |
| `registry.access.redhat.com/ubi9`  | + <br /> Git not found |  - | [click here](https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com/#https://github.com/RomanNikitenko/test-redhat-nodejs/tree/ubi9-devspaces-jenkins)  |
[...]

PR #587, opened 2024-06-03
Migrate ds-code container from ubi to rhel-els

azatsarynnyy:
https://issues.redhat.com/browse/CRW-6354

Migrates ds-code container from ubi to rhel-els

### DIFF
--- a/devspaces-code/build/dockerfiles/brew.Dockerfile
+++ b/devspaces-code/build/dockerfiles/brew.Dockerfile
@@ -15,8 +15,8 @@
 ############################# BUILD 1: libc-ubi8 ########################
 #########################################################################
 
-# https://registry.access.redhat.com/ubi8/nodejs-18
-FROM ubi8/nodejs-18:1-110.1716503936 as checode-linux-libc-ubi8-builder
+# https://registry.access.redhat.com/rhel9-6-els/rhel
+FROM registry.redhat.io/rhel8-6-els/rhel:8.6 as checode-linux-libc-ubi8-builder
 # hadolint ignore=DL3002
 USER root
 
@@ -224,8 +224,8 @@ RUN yarn \
 ############################# BUILD 2: libc-ubi9 ########################
 #########################################################################
 
-# https://registry.access.redhat.com/ubi9/nodejs-18
-FROM ubi9/nodejs-18:1-108.1716477799 as checode-linux-libc-ubi9-builder
+# https://registry.access.redhat.com/rhel9-2-els/rhel
+FROM registry.redhat.io/rhel9-2-els/rhel:9.2-1222 as checode-linux-libc-ubi9-builder
 # hadolint ignore=DL3002
 USER root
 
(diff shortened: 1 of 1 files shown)


devspaces-code/build/dockerfiles/brew.Dockerfile
```diff
@@ -15,8 +15,8 @@
 ############################# BUILD 1: libc-ubi8 ########################
 #########################################################################
 
-# https://registry.access.redhat.com/ubi8/nodejs-18
-FROM ubi8/nodejs-18:1-110.1716503936 as checode-linux-libc-ubi8-builder
+# https://registry.access.redhat.com/rhel9-6-els/rhel
+FROM registry.redhat.io/rhel8-6-els/rhel:8.6 as checode-linux-libc-ubi8-builder
```
RomanNikitenko, line 19: should we provide similar changes to https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-code/build/dockerfiles/Dockerfile ?

SDawley: Looking at the sync.sh script the base Dockerfile is no longer used as a template for the brew one, so it might not be needed at all.

It also looks like the sync script already ignores the build/dockerfiles directory so we don't have to worry about these changes getting overwritten.

SDawley: It will probably get updated automatically, but we should specify a build number, not just 8.6, for reproducibility etc

devspaces-code/build/dockerfiles/brew.Dockerfile
```diff
@@ -457,15 +457,16 @@ USER root
...
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -a -ldflags '-w -s' -a -installsuffix cgo -o che-machine-exec . && \
+RUN dnf -y install golang && \
+    CGO_ENABLED=0 GOOS=linux go build -mod=vendor -a -ldflags '-w -s' -a -installsuffix cgo -o che-machine-exec . && \
     mkdir -p /rootfs/go/bin && cp -rf $REMOTE_SOURCES_DIR/devspaces-images-code/app/devspaces-machineexec/che-machine-exec /rootfs/go/bin
 
 #########################################################################
 ############################# BUILD 4: rootfs assembly ##################
 #########################################################################
 
 # https://registry.access.redhat.com/ubi8
```
SDawley, line 468: This comment should be updated as well, its how the updateBaseImages script knows what to look for.

SDawley:
This should definitely be tested in brew first to make sure there aren't any more gaps between the original source images and rhel-els

SDawley:
In addition the https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-code/content_sets.yml file will need to be updated to use the els repos.

[https://github.com/redhat-developer/devspaces-images/commit/f2b6a52a76144fc524a7f916f88008ad13948277#diff-dc57e5f4a721821efe63d9a0a3e5760f746ee2ae6460af72cab65b8010baa634](Example)

So lines 255-257 will need to be updated as well 
https://github.com/redhat-developer/devspaces-images/blob/0d28a2b94ab0a321f77218947c323f42fd1f1b51/devspaces-code/build/dockerfiles/brew.Dockerfile#L255-L257

devspaces-code/build/dockerfiles/brew.Dockerfile
RomanNikitenko: - my comment was not about build on brew
- we had that Dockerfile for local build
-  some time ago I found that it was broken, so I fixed it within https://github.com/redhat-developer/devspaces-images/pull/547

It allows to build devspaces-code locally and it's helpful for investigation a problem(not related to brew) - as an investigation/debugging on brew is not easy process.

SDawley: Ah OK.

PR #592, opened 2024-07-01
Migrate Devspaces-Code to Node.js 20

RomanNikitenko:
Adapt Devspaces-Code to Node.js 20.
Issue: https://issues.redhat.com/browse/CRW-6620


vitaliy-guliy (approved):
Haven't tested it, but code looks well.
It will be great if we have some PR checks is possible.

PR #619, opened 2024-08-29
Fix link to the official documentation

dmytro-ndp:
There is still incorrect link in the "About" menu: `https://docs.redhat.com/**en-us**/documentation/red_hat_openshift_dev_spaces/3.16` , which is leading to 404 page error.

Correct one: `https://docs.redhat.com/*en*/documentation/red_hat_openshift_dev_spaces/3.16`


SDawley (approved):
lgtm

PR #624, opened 2024-08-30
Update Dashboard to use els image and fix build issue with yarn 3 update

SDawley:
https://issues.redhat.com/browse/CRW-6054
Updating dashboard to use rhel-els image. Installing g++ is necessary for build to run on s390x.

Fixing link to yarn since the new file has extension cjs and this wasn't caught in testing because of how cachito grabs files.


olexii4 (approved):
LGTM

svor (approved):

PR #626, opened 2024-09-03
Fix name of yarn file in dashboard directory for Traefik and Code

SDawley:
The extension on the yarn file changed from .js to .cjs so other builds dependent on that file need to be updated.


svor (approved):
lgtm

svor (approved):
lgtm

PR #633, opened 2024-09-16
CRW-7151 Removing devfileregistry container and all references to it.

SDawley:
Requesting review from Anatolii before merging so I don't accidentally remove something that's still needed for the new devfile implementation.


tolusha:
I am afraid but we need those files

devspaces-operator-bundle/build/scripts/insert-related-images-to-csv.sh
devspaces-operator-bundle/build/scripts/images.sh

SDawley:
@tolusha What do we need insert-related-images-to-csv for? Everything but the sed statements on lines 119 to 125 directly referenced the devfileregistry container, and I moved the sed lines to sync-che-olm where the script was called (lines 466-475).

I couldn't find where images.sh is called but I can add it back.

tolusha:
I can find niether.
So, let's remove those files.

tolusha (approved):

PR #647, opened 2024-10-09
Remove midstream infra related to the resources.tgz file that was removed upstream.

SDawley:
Should help address CRW-7481 and CRW-7482


svor (commented):
We still need ovsx  dockerfile and the build script
I created another PR to fix the build: https://github.com/redhat-developer/devspaces-images/pull/649

PR #662, opened 2024-12-05
chore: Adapt vscode/ripgrep version detection to npm build

RomanNikitenko:
Issue:  https://issues.redhat.com/browse/CRW-7866

There is a problem with `vscode/ripgrep` version detection after moving Dev Spaces Code to the `npm` build.
`npm list` works fine if `npm install` was executed and dependencies are already installed.
But there is no installed packages on the step we use that command . 
Thanks @SDawley for the investigation!

Within the current PR I propose to change a way how `vscode/ripgrep` version is detected.

rh-pre-commit.version: 2.2.0


SDawley (approved):
Thank you for the quick fix!

RomanNikitenko:
> Thank you for the quick fix!

Hope it will help :crossed_fingers:

PR #693, opened 2024-12-11
CRW-7881 Update Idea to build as rhel9

SDawley:
Keep idea consistent with other containers.
https://issues.redhat.com/browse/CRW-7881


SDawley:
@tolusha I'm unsure about the bundle changes, specifically to the idea injector, would you please take a look?

PR #697, opened 2024-12-18
chore: Use x86_64 platform only for devspaces-code build

RomanNikitenko:
Currently `devspaces-code` build is failing.
Can we temporary turn off build for the `ppc64le` and `s390x` platforms?

I would like to get it working for the `x86_64` platform first
then
I can investigate problems specific for `ppc64le` and `s390x`

rh-pre-commit.version: 2.2.0


SDawley:
@RomanNikitenko This file configures the downstream builds and wouldn't fix any of the issues we're having in Jenkins.

PR #703, opened 2024-12-19
Add upstream UBI/UDI bug fixes

dkwon17:
This PR adds UBI/UDI bug fixes from upstream. They include:

* adds podman 5 support: https://github.com/devfile/developer-images/pull/209
* updates .stow-local-ignore with files in /home/user that would break running stow: https://github.com/devfile/developer-images/pull/196
* automatically detects fuse/vfs: https://github.com/devfile/developer-images/pull/198


dkwon17:
This PR was replaced by https://github.com/redhat-developer/devspaces-images/pull/729

PR #714, opened 2025-02-04
fix: Ability to start code editor in a container without libbrotli lib

RomanNikitenko:
Cache `libbrotli` from the builder image to provide it to a user's container.
Fixes: https://issues.redhat.com/browse/CRW-7975


SDawley (approved):
lgtm

PR #760, opened 2025-04-18
fix: Do not download Node.js when build code on brew

RomanNikitenko:
At the moment devspaces-code fails on brew when it tries to download Node.js binary, see https://issues.redhat.com/browse/CRW-8614
Actually, the binary is copied from the build-image and should not be downloaded.

The current change should fix the problem.


RomanNikitenko:
@SDawley 
yes, you can look at https://jenkins-csb-crwqe-main.dno.corp.redhat.com/job/DS_CI/job/get-sources-rhpkg-container-build_3.x/9340/consoleFull
The build was successful.

At the end of logs you can find the dockerfile content that was used for the build.
- put `RUN sed -i -r -e '/function npmInstall/ !s|^[^#]*npmInstal|//&|' build/npm/postinstall.js` as a search line
- it was the previous line for the deleted section in the PR

> make sure we don't need any of those properties?

Actually those properties come from the upstream (che-code).
Cachito used to clear all project's '.npmrc' files that's why there was that workaround.
It looks like now it's not as I found double similar lines in the `.npmrc`(for example, two target="30.5.1 and so on).
So, I think we don't need that workaround anymore.

SDawley (approved):

PR #771, opened 2025-05-05
Reduce memory when compile devspaces-code

RomanNikitenko:
See https://issues.redhat.com/browse/CRW-8646?focusedId=27115737&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27115737

PR #777, opened 2025-05-13
fix: Reduce memory when compile devspaces-code assembly

RomanNikitenko:
See https://github.com/redhat-developer/devspaces-images/pull/771


SDawley (approved):
lgtm

